PR #1, opened 2021-06-07
list images

ashnamehrotra:


go.mod
```diff
@@ -0,0 +1,9 @@
+module os/exec
```
sozercan: module name needs to be "github.com/azure/eraser"

sozercan:
no need to check in the binary

sozercan:
let's add a Makefile as a follow up. this will make it easier to build images and push
example: https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/Makefile

go.mod
sozercan: you should run `go mod tidy` after

pod.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: list_images
+  labels:
+    name: list_images
+spec:
+  containers:
+  - name: list_images
+    image: ashnam/list_images
+    ports:
+    - containerPort: 80
```
sozercan: nit: add a newline

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -ldflags '-w -extldflags "-static"' -o /out/example .
+
+RUN chmod +x /out/example
+
+FROM scratch as bin
+
+COPY --from=build /out/example /
+
+ENTRYPOINT ["/example"]
```
sozercan: nit: add a newline

eraser.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/containerd/containerd"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+)
+
+func get_containerd_images() error {
```
sozercan: Go convention is to use mixed caps: https://golang.org/doc/effective_go#mixed-caps instead of underscores

go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/azure/eraser
```
sozercan:
```suggestion
module github.com/Azure/eraser
```

sozercan:
can you add a .gitignore to ignore the binary?

sozercan (approved):

PR #17, opened 2021-07-09
Initial implementation of CRDs and controllers

### LINKED ISSUE #8: create CRDs
Create CRDs for ImageJob and ImageList

### LINKED ISSUE #12: Add eraser removal functionality by image name
Add eraser removal functionality by name, not just sha id as it does currently

### LINKED ISSUE #15: Fix file structure
Should not be 2 folders for imagejob and imagelist

ashnamehrotra:


config/crd/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
+# This kustomization.yaml is not intended to be run by itself,
+# since it depends on service name and namespace that are out of this kustomize package.
+# It should be run by config/default
+resources:
+- bases/eraser.sh_imagelists.yaml
+#+kubebuilder:scaffold:crdkustomizeresource
```
sozercan, line 8: need other crds here too

sozercan: make sure all deploys using `make deploy`

api/v1alpha1/imagestatus_types.go
```diff
@@ -0,0 +1,65 @@
+/*
```
sozercan, line 1: rename CRD `ImageJobStatus`

mylist.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: eraser.sh/v1alpha1
```
sozercan: we don't need this file

controllers/imagelist/imagelist_controller.go
```diff
@@ -0,0 +1,110 @@
...
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
+func (r *ImageListReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	// your logic here
+	//log.Print("")
+	fmt.Print("hello world")
```
sozercan: change this to controller-runtime logger, see `setupLog` in main for an example

Dockerfile
```diff
@@ -1,17 +1,27 @@
-FROM golang:1.16 AS build
```
sozercan, line 1: we still want this dockerfile since this is for removing images. You can name it `Dockerfile.eraser` or something to differentiate

sozercan: and obviously change `example`

config/samples/eraser_v1alpha1_imagejob.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: imagejob-sample
+spec:
+   jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: eraser
+            image: ashnam/list-images
```
sozercan: let's update `list-images` name

Makefile
```diff
@@ -1,5 +1,108 @@
...
+
+run: manifests generate fmt vet ## Run a controller from your host.
+	go run ./main.go
+
+docker-build: test ## Build docker image with the manager.
+	docker build -t ${IMG} .
+
+docker-push: ## Push docker image with the manager.
+	docker push ${IMG}
+
```
sozercan, line 71: add `docker-build-eraser`/`docker-push-eraser` or something to build the eraser image (not the manager but the image that lists and removes images)

config/samples/eraser_v1alpha1_imagestatus.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: eraser.sh/v1alpha1
+kind: ImageStatus
```
sozercan, line 2: same comment re: `ImageJobStatus`

Makefile
```diff
@@ -1,5 +1,108 @@
...
+##@ Development
+
+manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+
+generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+
+fmt: ## Run go fmt against code.
+	go fmt ./...
```
cpuguy83, line 47: This will just dump the go fmt results to stdout and not actually fmt the codebase, not sure if that is what was intended or not.

Makefile
```diff
@@ -1,5 +1,108 @@
...
+controller-gen: ## Download controller-gen locally if necessary.
+	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1)
+
+KUSTOMIZE = $(shell pwd)/bin/kustomize
+kustomize: ## Download kustomize locally if necessary.
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
+
+# go-get-tool will 'go get' any package $2 and install it to $1.
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+define go-get-tool
```
cpuguy83, line 103: `go install <module>` will accomplish this.

api/v1alpha1/imagejob_types.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+type JobSpec struct {
```
cpuguy83: What does this do?

api/v1alpha1/imagelist_types.go
```diff
@@ -0,0 +1,71 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ImageListSpec   `json:"spec,omitempty"`
+	Status ImageListStatus `json:"status,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+
+// ImageListList contains a list of ImageList
+type ImageListList struct {
```
cpuguy83, line 63: 👀

Dockerfile
sozercan: @ashnamehrotra can you make sure to add the Dockerfile for eraser

config/samples/eraser_v1alpha1_imagejob.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: imagejob-sample
+spec:
+   jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: eraser
+            image: ashnam/remove-images
```
sozercan: can you add a tag here too

Makefile
```diff
@@ -1,5 +1,115 @@
...
+
+docker-build: test ## Build docker image with the manager.
+	docker build -t ${IMG} .
+
+docker-push: ## Push docker image with the manager.
+	docker push ${IMG}
+
+docker-build-eraser:
+	cd pkg/eraser
+	docker build -t ashnam/remove_images .
```
sozercan: add tags here

Makefile
```diff
@@ -1,5 +1,115 @@
...
+docker-push-eraser:
+	cd pkg/eraser
+	docker push ashnam/remove_images
```
sozercan: add tags here

Makefile
```diff
@@ -1,5 +1,115 @@
...
+	go run ./main.go
+
+docker-build: test ## Build docker image with the manager.
+	docker build -t ${IMG} .
+
+docker-push: ## Push docker image with the manager.
+	docker push ${IMG}
+
+docker-build-eraser:
+	cd pkg/eraser
```
sozercan: i don't see pkg/eraser, you can also just use `-f /path/to/dockerfile` instead of cd

Makefile
sozercan: I think this should format/replace too

Makefile
sozercan: this is part of kubebuilder Makefile. Can we skip this?

api/v1alpha1/imagejob_types.go
sozercan: looks like this is removed now. @ashnamehrotra can you comment/resolve when it's resolved? thanks!

Makefile
cpuguy83: Oh... `go fmt`, not `gofmt`.

api/v1alpha1/imagejob_types.go
ashnamehrotra: yes it was old code that should have been removed

Makefile
ashnamehrotra: pkg should be under the main eraser folder, and we moved the removal code from before under a different eraser folder (pkg/eraser)

Dockerfile
ashnamehrotra: should be under pkg/eraser as Dockerfile.eraser

pkg/eraser/eraser.go
```diff
@@ -0,0 +1,248 @@
...
+func main() {
+
+	// image job can pass the imagelist into each pod as a env variable maybe
+
+	ctx := cli.NewContext(nil, nil, nil)
+
+	imageClient, conn, err := getImageClient(ctx)
+
+	if err != nil {
+		fmt.Printf("image client err")
```
sozercan: let's handle all these errors (applies to all)

pkg/eraser/eraser.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: remove-images
+  labels:
+    name: remove-images
+spec:
+  containers:
+  - name: remove-images
+    image: ashnam/remove_images
```
sozercan: tag here

Makefile
sozercan: we don't need to cd here. we can just push

sozercan: for build, we can just do `docker build -t ... -f pkg/eraser/Dockerfile`

pkg/eraser/eraser.go
```diff
@@ -0,0 +1,248 @@
...
+	for _, entry := range intSlice {
+		if _, value := keys[entry]; !value {
+			keys[entry] = true
+			list = append(list, entry)
+		}
+	}
+	return list
+}
+
+func main() {
```
cpuguy83: Let's move all the app logic out of `main()`.
I tend to keep `main()` for setting up things related to setting up the process: flags, log config, and general bootstrapping. Then we can move app logic to something that returns an `error` which we can handle in `main()` (such as print to stderr and exit non-zero)

pkg/eraser/eraser.go
```diff
@@ -0,0 +1,248 @@
...
+		m[img.Id] = img.RepoTags
+	}
+
+	response, err := pb.NewRuntimeServiceClient(conn).ListContainers(context.Background(), new(pb.ListContainersRequest))
+
+	if err != nil {
+		fmt.Printf("list containers err")
+	}
+
+	var runningImages []string
```
cpuguy83: You can make this a `map[string]struct{}` and it will automatically de-up as you insert and provides an index so you don't need to iterate the list to determine if an image is in the list.

cpuguy83 (changes requested):

cpuguy83:
Maybe make sure we have the bin name in .gitignore and before merge let's make sure to squash so we don't have that binary in the intermediate commits.

pkg/eraser/eraser.go
```diff
@@ -0,0 +1,248 @@
...
+
+	imageClient := pb.NewImageServiceClient(conn)
+
+	return imageClient, conn, nil
+}
+
+func ListImages(client pb.ImageServiceClient, image string) (resp *pb.ListImagesResponse, err error) {
+	request := &pb.ListImagesRequest{Filter: &pb.ImageFilter{Image: &pb.ImageSpec{Image: image}}}
+
+	resp, err = client.ListImages(context.Background(), request)
```
cpuguy83: Let's make sure to plumb a context through instead of using `context.Background()`.
We'll also likely want to set timeouts on the context.

pkg/eraser/eraser.go
ashnamehrotra: what would the struct be?

cpuguy83: In this case we are just using the map for a lookup value, so this would either be an empty struct (`struct{}`) where we throw away the value. Example: https://play.golang.org/p/DI1I08rZRom

So we'd use this like so:

```go
for _, img := range allImages {
    if _, isRunning := runningImages[img]; !isRunning {
        nonRunningImages = append(nonRunningImages, img)
    }
}
```

We are using the map to tell us if the image is in the running list rather than iterating over all the objects in `runningImages`

ashnamehrotra: got it thank you!

pkg/eraser/eraser.go
```diff
@@ -89,153 +92,121 @@ func getImageClient(context *cli.Context) (pb.ImageServiceClient, *grpc.ClientCo
 	return imageClient, conn, nil
 }
 
-func ListImages(client pb.ImageServiceClient, image string) (resp *pb.ListImagesResponse, err error) {
+func ListImages(client pb.ImageServiceClient, image string, ctx context.Context) (resp *pb.ListImagesResponse, err error) {
```
sozercan: no need to export this

move ctx as first arg

pkg/eraser/eraser.go
```diff
@@ -89,153 +92,121 @@ func getImageClient(context *cli.Context) (pb.ImageServiceClient, *grpc.ClientCo
...
+	if err != nil {
+		return nil, err
+	}
 
-	return
+	return resp, nil
 }
 
-func RemoveImage(client pb.ImageServiceClient, image string) (resp *pb.RemoveImageResponse, err error) {
+func RemoveImage(client pb.ImageServiceClient, image string, ctx context.Context) (resp *pb.RemoveImageResponse, err error) {
```
sozercan: no need to export this

move ctx as first arg

cpuguy83:
Looks like you have a merge conflict.

sozercan (approved):
LGTM

controllers/controller.go
```diff
@@ -0,0 +1,29 @@
...
+func init() {
+	controllerAddFuncs = append(controllerAddFuncs, imagelist.Add)
+	controllerAddFuncs = append(controllerAddFuncs, imagejob.Add)
+}
+
+func SetupWithManager(m manager.Manager) error {
+	for _, f := range controllerAddFuncs {
+		if err := f(m); err != nil {
+			if kindMatchErr, ok := err.(*meta.NoKindMatchError); ok {
+				klog.Infof("CRD %v is not installed, its controller will perform noops!", kindMatchErr.GroupKind)
```
sozercan: change this to controller-runtime's logger

pkg/eraser/eraser.go
```diff
@@ -118,104 +114,129 @@ func RemoveImage(ctx context.Context, client pb.ImageServiceClient, image string
...
+	runningImages := make(map[string]struct{})
 
 	for _, container := range response.Containers {
 		curr := container.Image
-		runningImages = append(runningImages, curr.GetImage())
+		runningImages[curr.GetImage()] = struct{}{}
 	}
 
-	var nonRunningImages []string
+	nonRunningImages := make(map[string]struct{})
```
cpuguy83: No need to change for here, but on these map creations, since we already know the size we need, we can pre-allocate the map with `make(map[string]struct{}, len(allImages)-len(runningImages))`

cpuguy83 (approved):
LGTM, good next step.

cpuguy83:
Another merge conflict on eraser.go

pkg/eraser/eraser.go
sozercan: @AldairCoronel just added the `prealloc` linter, which should make it easy to point these out

PR #20, opened 2021-07-14
Add CI for lint and unit tests

### LINKED ISSUE #3: add CI for lint and unit tests
add a CI using [github actions](https://github.com/Azure/eraser/actions/new) for basic checks

- [ ] linter - [golangci-lint action](https://github.com/golangci/golangci-lint-action)
- [ ] run unit tests

AldairCoronel:
Added the workflow.yaml


.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,35 @@
+name: build_test
```
sozercan, line 1:
```suggestion
name: test
```

we are not building yet, let's just change to test

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,35 @@
...
+#     steps:
+#       - name: Set up Go 1.16
+#         uses: actions/setup-go@v2
+#         with:
+#           go-version: 1.16
+
+#       - name: Check out code into the Go module directory
+#         uses: actions/checkout@v2
+
+#       - name: Unit test
```
sozercan, line 35: why is this commented?

AldairCoronel: we don't have unit tests yet, right?

sozercan: that's okay. it'll just pick it up once we have it. let's uncomment and add that

PR #27, opened 2021-07-23
Eraser.go changes for test

ashnamehrotra:
created client interface that has listImages, listContainers, and removeImage


pkg/eraser/eraser.yaml
```diff
@@ -16,6 +16,7 @@ spec:
     ports:
     - containerPort: 80
     imagePullPolicy: Always
+    args:  ["--runtime=containerd", "imagelist=imagelist-sample"]
```
sozercan, line 19: why is this hardcoded?

ashnamehrotra: the eraser.yaml isn't used in the project, it was just for testing deploying the pods manually

sozercan: aren't we automatically setting runtime value

pkg/eraser/eraser.go
```diff
@@ -223,14 +233,40 @@ func removeVulnerableImages() (err error) {
...
-	err := removeVulnerableImages()
 
+	runtimePtr := flag.String("runtime", "containerd", "container runtime")
+	imageListPtr := flag.String("imagelist", "", "name of ImageList")
+
+	flag.Parse()
+
+	var socketPath string
+
+	// switch case
```
sozercan, line 244: let's make this a switch

sozercan:
can we focus this on changes to eraser.go only? let's rebase to master

pkg/eraser/eraser.yaml
ashnamehrotra: yeah but only when imagejob schedules the pods . When scheduling the pods manually for testing purposes we have to provide it

PR #31, opened 2021-07-27
Add Unit Tests for Eraser 

### LINKED ISSUE #22: add unit tests for eraser

AldairCoronel:
Fixes #22


pkg/eraser/eraser.go
```diff
@@ -22,16 +23,67 @@ const (
 
 var (
 	// Timeout  of connecting to server (default: 10s)
-	timeout time.Duration = 100 * time.Second
+	timeout = 10 * time.Second
 )
 
+var ErrProtocolNotSupported = errors.New("protocol not supported")
```
sozercan: Add these into existing var block

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,219 @@
+package main
```
sozercan: looks like this is not picked up by `make test`?

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	image5 := pb.Image{
+		Id:          "sha256:fd46ec1af6de89db1714a243efa1e35c4408f5a5b9df9c653dd70db1ee95522b",
+		RepoDigests: []string{"docker.io/ashnam/remove_images@sha256:d93d3d3073797258ef06c39e2dce9782c5c8a2315359337448e140c14423928e"},
+	}
+
+	// containers
+	container1 := pb.Container{
+		Id:           "7eb07fbb43e86a6114fb3b382339176117bc377cff89d5466210cbf2b101d4cb",
+		PodSandboxId: "07adb7d1a87e7f79490f29846658f90456709c2b754dcffacad843d87033b2ef",
```
sozercan: let's cleanup these and only have the ones we are actually using

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	for _, tc := range testCases {
+		a, _, e := GetAddressAndDialer(tc.endpoint)
+		if a != tc.addr || !errors.Is(e, tc.err) {
+			t.Errorf("Test fails")
+		}
+	}
+}
+
+var ErrImageNotRemoved = errors.New("image not removed")
```
sozercan: use a var block instead of individual vars

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,219 @@
...
+		Labels:       map[string]string{"io.kubernetes.container.name": "kube-proxy", "io.kubernetes.pod.name": "kube-proxy-s5c67", "io.kubernetes.pod.namespace": "kube-system", "io.kubernetes.pod.uid": "f6dddad9-b772-4735-a8df-db910fdca461"},
+		Annotations:  map[string]string{"io.kubernetes.container.hash": "a11f949b", "io.kubernetes.container.restartCount": "0", "io.kubernetes.container.terminationMessagePath": "/dev/termination-log", "io.kubernetes.container.terminationMessagePolicy": "File", "io.kubernetes.pod.terminationGracePeriod": "302"},
+	}
+
+	client := testClient{
+		containers: []*pb.Container{&container1, &container2},
+		images:     []*pb.Image{&image1, &image2, &image3, &image4, &image5},
+	}
+
+	removeVulnerableImages(&client, "", "")
```
sozercan: where are we checking expected output?

sozercan: we should validate if it runs an error or not

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+var ErrImageNotRemoved = errors.New("image not removed")
+var ErrImageEmpty = errors.New("unable to remove empty image")
+
+type testClient struct {
+	containers []*pb.Container
+	images     []*pb.Image
+}
+
+func (client *testClient) listImages(ctx context.Context) (list []*pb.Image, err error) {
```
sozercan: let's use `c` here for consistency

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,330 @@
...
+func TestListImages(t *testing.T) {
+	var testCases = []struct {
+		imagesInput  testClient
+		imagesOutput []*pb.Image
+		err          error
+	}{
+		{
+			imagesInput: testClient{
+				containers: []*pb.Container{},
+				images:     []*pb.Image{&image1, &image2, &image3, &image4, &image5},
```
sozercan, line 246: let's add more edge cases. examples: 0->0, 1->0

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+}
+
+func TestRemoveImage(t *testing.T) {
+	var testCases = []struct {
+		imagesInput   testClient
+		imageToDelete string
+		imagesOutput  []*pb.Image
+		err           error
+	}{
```
sozercan, line 381: let's add more edge cases. examples: 0->0, 1->0

pkg/eraser/eraser_test.go
sozercan, line 304: add more edge cases. example: 0->0

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func TestListContainers(t *testing.T) {
+	var testCases = []struct {
+		containersInput testClient
+		containerOutput []*pb.Container
+		err             error
+	}{
+		{
+			containersInput: testClient{
```
sozercan, line 338: more test cases here too

sozercan:
looks like there's a conflict

config/rbac/role.yaml
```diff
@@ -6,6 +6,26 @@ metadata:
   creationTimestamp: null
   name: manager-role
 rules:
+- apiGroups:
```
sozercan: why changes here?

pkg/eraser/eraser_test.go
sozercan: delete with a image name (repotags), delete with image name@sha, etc

config/rbac/role.yaml
AldairCoronel: I pulled one of Ashna's branches (to get eraser.go changes) and mistakenly committed that file. I already reversed it.

pkg/eraser/eraser_test.go
AldairCoronel: At the moment it is not possible to do unit tests of this function. I removed it and added unit tests for the methods that implement the Client interface in eraser_test.

AldairCoronel: At the moment you cannot write unit test of that function. I removed it and added unit tests for the methods that implement the Client interface in eraser_test.

go.mod
```diff
@@ -5,9 +5,10 @@ go 1.13
 require (
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.13.0
+	google.golang.org/grpc v1.27.1
 	k8s.io/api v0.21.2
 	k8s.io/apimachinery v0.21.2
 	k8s.io/client-go v0.21.2
-	k8s.io/klog v1.0.0
+	k8s.io/cri-api v0.21.3
```
sozercan: can you run `go mod tidy`? these should go away i think

pkg/eraser/eraser.go
```diff
@@ -189,15 +198,41 @@ func removeVulnerableImages(c Client, socketPath string, imagelistName string) (
...
+	result := eraserv1alpha1.ImageList{}
+	err = clientset.RESTClient().Get().
+		AbsPath("apis/eraser.sh/v1alpha1").
+		Namespace("eraser-system").
+		Resource("imagelists").
+		Name(imagelistName).
+		Do(backgroundContext).Into(&result)
+
+	if err != nil {
+		log.Println("unable to find imagelist")
```
sozercan, line 221: add more info here please

pkg/eraser/eraser.go
```diff
@@ -189,15 +198,41 @@ func removeVulnerableImages(c Client, socketPath string, imagelistName string) (
...
+	if err != nil {
+		return err
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	result := eraserv1alpha1.ImageList{}
```
sozercan: i think we can revert these changes, as other pr adds them

sozercan: do current unit tests touch these?

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,382 @@
...
+			containerOutput: []*pb.Container{&container1, &container2},
+			err:             nil,
+		},
+		{
+			containersInput: testClient{
+				containers: []*pb.Container{},
+				images:     []*pb.Image{},
+			},
+			containerOutput: []*pb.Container{},
+			err:             nil,
```
sozercan, line 351: what happens with no container but with images? and opposite

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,382 @@
...
+		},
+		{
+			imagesInput: testClient{
+				containers: []*pb.Container{},
+				images:     []*pb.Image{&image1},
+			},
+			imageToDelete: "sha256:ccd78eb0f420877b5513f61bf470dd379d8e8672671115d65c6f69d1c4261f87",
+			imagesOutput:  []*pb.Image{},
+			err:           nil,
+		},
```
sozercan, line 435: another case where imageToDelete doesn't exist in our list

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,382 @@
...
+	for _, tc := range testCases {
+		l, e := tc.containersInput.listContainers(backgroundContext)
+		if testEqContainers(l, tc.containerOutput) == false || !errors.Is(e, tc.err) {
+			t.Errorf("Test fails")
+		}
+	}
+
+}
+
+func TestRemoveImage(t *testing.T) {
```
sozercan, line 375: can we add a test for running images too? try to delete but it shouldn't delete

pkg/eraser/eraser.go
AldairCoronel: No, I only need the inferface above. I reverted these changes.

pkg/eraser/eraser_test.go
AldairCoronel: Yes! I added it. Before I was not checking if the image was running before removing so I added that functionality and changed the unit tests.

sozercan (approved):
LGTM

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,412 @@
...
+			protocol:         "unix",
+			addr:             "",
+			err:              nil,
+		},
+		{
+			endpoint:         "tcp://localhost:8080",
+			fallbackProtocol: "unix",
+			protocol:         "tcp",
+			addr:             "localhost:8080",
+			err:              nil,
```
cpuguy83: Do we have any negative cases where we expect an error?

pkg/eraser/eraser_test.go
```diff
@@ -0,0 +1,412 @@
...
+			containersInput: testClient{
+				containers: []*pb.Container{},
+				images:     []*pb.Image{&image1},
+			},
+			containerOutput: []*pb.Container{},
+			err:             nil,
+		},
+	}
+
+	backgroundContext, cancel := context.WithTimeout(context.Background(), timeoutTest)
```
cpuguy83: `ctx` is the idiomatic name for this.

cpuguy83 (changes requested):

sozercan:
@AldairCoronel looks like some conflicts now

pkg/eraser/eraser_test.go
```diff
@@ -43,13 +44,22 @@ func TestParseEndpointWithFallBackProtocol(t *testing.T) {
 			fallbackProtocol: "unix",
 			protocol:         "",
 			addr:             "",
-			err:              ErrParsing,
+			err:              &url.Error{},
```
cpuguy83: WDYT if instead oof a literal error we do something like passing a `func(t *testing.T, err error)` to do the error check.
So in the test cases the `err` field would be a closure, and in the actual test code below we pass it like `tc.errCheck(t, err)`.

I think that gets rid of this weird case with url.Error, and we can be more specific with the check.

pkg/eraser/eraser_test.go
```diff
@@ -85,12 +95,21 @@ func TestParseEndpoint(t *testing.T) {
...
 	for _, tc := range testCases {
 		p, a, e := parseEndpoint(tc.endpoint)
-		if p != tc.protocol || a != tc.addr || !errors.Is(e, tc.err) {
+
+		as := tc.err
+		var errAux bool
+		if as == nil {
+			errAux = !errors.Is(e, as)
+		} else {
+			errAux = !errors.As(e, &as)
```
cpuguy83: I don't think this is doing what we wanted.
As an example: https://play.golang.org/p/a4qR3ervYOw.

There doesn't really seem to be a good way to assert the url.Error and also other errors we want to catch with the `errors` package without having the literal error that `url.Error` is wrapping (in which case we could use `errors.Is`).
Hence, my [suggestion](https://github.com/Azure/eraser/pull/31/files/5b1ed2b6f327cb77659991e5eeeb80e4de0a2c3a..a33cf0387adc0a25e947a09e215c2d9321496dcc#r681924378).

cpuguy83 (changes requested):

cpuguy83 (approved):
LGTM

PR #45, opened 2021-08-04
Change registry

### LINKED ISSUE #23: replace personal registry
replace `ashnam` docker hub registry with another registry (like [GHCR](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry))

ashnamehrotra:


.github/workflows/build-eraser.yaml
```diff
@@ -0,0 +1,43 @@
+name: Build and Push Eraser Image
```
sozercan: let's call this manager (change filename too)

controllers/imagelist/imagelist_controller.go
```diff
@@ -90,8 +90,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 						RestartPolicy: "Never",
 						Containers: []corev1.Container{
 							{
-								Name:            "remove-images",
-								Image:           "aldaircoronel/remove_images:latest",
+								Name:            "eraser",
+								Image:           "ghcr.io/azure/eraser:latest",
```
sozercan, line 94: this doesn't exist. @AldairCoronel with the e2e PR we should be able to override this.

sozercan (approved):
LGTM

Makefile
```diff
@@ -1,6 +1,7 @@
 
 # Image URL to use all building/pushing image targets
-IMG ?= aldaircoronel/controller:latest
+IMG ?= ghcr.io/azure/eraser-manager:latest
```
cpuguy83, line 3: Did we want this to be something other than latest?

```suggestion
COMMIT := $(shell git rev-parse --short HEAD)
DIRTY := $(shell [ -n "$(git status --porcelain)" ] && echo "-dirty")
IMG ?= ghcr.io/azure/eraser-manager:$(COMMIT)$(DIRTY)
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Log in to the GHCR
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publish
+        run: |
+          make docker-push IMG=${{ env.REGISTRY }}/${{ env.MANAGER_IMAGE_REPO }}:${TAG}
```
cpuguy83, line 69: Note that you can add `--push` to the build command.

cpuguy83 (approved):
Minor comment on image naming, otherwise LGTM

Makefile
sozercan: this will get overridden with `IMG` value (including tag) when a release is tagged

.github/workflows/release.yaml
sozercan: Yes, I wanted to make sure both images are built before pushing so if one of them fails to build we fail

cpuguy83: Ah yes that makes sense.

Makefile
sozercan: example: https://github.com/sozercan/eraser/runs/3435922240?check_suite_focus=true

PR #56, opened 2021-08-13
Cache all the things

cpuguy83:
This adds cache mounts for the go build cach which makes repeated
builds, even with changes to the source, use Go's caching mechanism.

Also adds builds to the CI to ensure Dockerfile's aren't broken.
This uses github actions as a cache backend for buildkit.

Signed-off-by: Brian Goff <cpuguy83@gmail.com>


cpuguy83:
Fresh unit test run in CI went from 3mins to 26s cached. 🎉

sozercan (approved):
fancy!

PR #85, opened 2022-02-16
Do not use extra module for pkg/eraser

cpuguy83:

    Multi-module repos are highly problematic for go currently and
    complicate development.
    Tools like gopls do not support it breaks the normal dev workflow.
    
    Instead just keep everything in one module for now.
    We may return to multi-module once the repo is stabilized or go is in a
    better place for multi-module development.

Fixes #


Dockerfile
```diff
@@ -1,29 +1,47 @@
...
-# Build
+FROM builder AS eraser-build
 RUN \
-    --mount=type=cache,target=/root/gocache \
-    GOCACHE=/root/gocache CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
+    --mount=type=cache,target=${GOCACHE} \
+    --mount=type=cache,target=/go/pkg/mod \
+    GOOS=linux GOARCH=amd64 go build -ldflags '-w -extldflags "-static"' -o out/removal ./pkg/eraser
+
+FROM scratch as eraser
```
sozercan: can we just use distroless/static:nonroot here? so we don't have to worry about updating/copying certs

Dockerfile
```diff
@@ -1,29 +1,47 @@
...
+COPY --from=eraser-build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=eraser-build /workspace/out/removal /
+ENTRYPOINT ["/removal"]
```
sozercan: should we rename this to eraser?

pkg/eraser/Dockerfile
```diff
@@ -1,22 +1,23 @@
-FROM golang:1.17 AS build
+# syntax=mcr.microsoft.com/oss/moby/dockerfile:1.3.1
```
sozercan: do we still need this file?

cpuguy83: I could have sworn I deleted this one, guess I missed it

Dockerfile
```diff
@@ -1,29 +1,47 @@
...
 WORKDIR /
-COPY --from=builder /workspace/manager .
+COPY --from=manager-build /workspace/out/manager .
 USER 65532:65532
-
 ENTRYPOINT ["/manager"]
+
+FROM scratch AS bins
+COPY --from=manager-build /workspace/out/manager /
+COPY --from=eraser /workspace/out/removal /
```
sozercan: nit: newline

sozercan:
not sure why cla is not green for you

Dockerfile
cpuguy83: Fixed

pkg/eraser/Dockerfile
cpuguy83: Fixed

Dockerfile
cpuguy83: Fixed

Dockerfile
```diff
@@ -1,29 +1,45 @@
...
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static:nonroot AS manager
 WORKDIR /
-COPY --from=builder /workspace/manager .
+COPY --from=manager-build /workspace/out/manager .
 USER 65532:65532
-
 ENTRYPOINT ["/manager"]
+
+FROM scratch AS bins
```
sozercan: what is this used for?

sozercan (approved):
comment above, otherwise LGTM. Thanks!

PR #99, opened 2022-02-23
imagelist as singleton

### LINKED ISSUE #98: Do multiple ImageList make sense?
Do we really need to have a CRD for imagelists?
Thinking about it, it seems strange that we'd maintain separate lists whereas a single config map could solve this problem.
To add an item to the list, update the config map. When the configmap is updated we can trigger a reconcile.

Using a config map also has the added benefit of giving us something we can just mount into the eraser pod as a volume for it to get the image list itself.

Managing multiple lists also seems a bit hairy to deal with.

sozercan:
Signed-off-by: Sertac Ozercan <sozercan@gmail.com>

Fixes #98 


controllers/imagelist/imagelist_controller.go
```diff
@@ -36,6 +37,7 @@ import (
 var (
 	log         = logf.Log.WithName("controller").WithValues("process", "imagelist-controller")
 	eraserImage = flag.String("eraser-image", "ghcr.io/azure/eraser:latest", "eraser image")
+	imageList   = types.NamespacedName{Name: "imagelist"}
```
sozercan, line 40: @ritazh @cpuguy83 fyi

sozercan: feel free to suggest other names

cpuguy83 (approved):
LGTM

PR #107, opened 2022-02-25
Remove imagestatus

### LINKED ISSUE #60: Refactor image list status
The imagelist status I think is going to be problematically large since it has the status for each node+image to be cleaned up.
I wonder if we can bubble up a summary into the imagelist status (maybe a copy of what's in imagejob) instead, with a link to the imagejob, and in imagejob link to each imagestatuses.

### LINKED ISSUE #96: rethink imagestatus?
Today we keep track of the result of an `imagejob` by creating an `imagestatus` for each node.
The simple version of this looks like:

```mermaid
graph TD;
ImageList1 --> ImageJob1;
ImageJob1 --> EraserPodJob1-Node1;
ImageJob1 --> EraserPodJob1-Node2;
ImageJob1 --> EraserPodJob1-NodeN;
EraserPodJob1-Node1 --> ImageStatus-Node1;
EraserPodJob1-Node2 --> ImageStatus-Node2;
EraserPodJob1-NodeN --> ImageStatus-NodeN;
ImageStatus-Node1 --> ImageJob1;
ImageStatus-Node2 --> ImageJob1;
ImageStatus-NodeN --> ImageJob1;
ImageJob1 --> ImageList1;
```

However if you add a 2nd `imagelist` to the mix, you may notice that `imagestatus` is per node, not per job per node.
This is significantly more complex and likely to cause issues:

```mermaid
graph TD;
ImageList1 --> ImageJob1;
ImageJob1 --> EraserPodJob1-Node1;
ImageJob1 --> EraserPodJob1-Node2;
ImageJob1 --> EraserPodJob1-NodeN;
EraserPodJob1-Node1 --> ImageStatus-Node1;
EraserPodJob1-Node2 --> ImageStatus-Node2;
EraserPodJob1-NodeN --> ImageStatus-NodeN;
ImageStatus-Node1 --> ImageJob1;
ImageStatus-Node2 --> ImageJob1;
ImageStatus-NodeN --> ImageJob1;
ImageJob1 --> ImageList1;
ImageList2 --> ImageJob2;
ImageJob2 --> EraserPodJob2-Node1;
ImageJob2 --> EraserPodJob2-Node2;
ImageJob2 --> EraserPodJob2-NodeN;
EraserPodJob2-Node1 --> ImageStatus-Node1;
EraserPodJob2-Node2 --> ImageStatus-Node2;
EraserPodJob2-NodeN --> ImageStatus-NodeN;
ImageStatus-Node1 --> ImageJob2;
ImageStatus-Node2 --> ImageJob2;
ImageStatus-NodeN --> ImageJob2;
ImageJob2 --> ImageList2;
```

Even if we do make this per job per node, this becomes significantly cleaner and less error prone, with each imagelist effectively creating a separate tree of events as one might expect:

```mermaid
graph TD;
ImageList1 --> ImageJob1;
ImageJob1 --> EraserPodJob1-Node1;
ImageJob1 --> EraserPodJob1-Node2;
ImageJob1 --> EraserPodJob1-NodeN;
EraserPodJob1-Node1 --> ImageStatusJob1-Node1;
EraserPodJob1-Node2 --> ImageStatusJob1-Node2;
EraserPodJob1-NodeN --> ImageStatusJob1-NodeN;
ImageStatusJob1-Node1 --> ImageJob1;
ImageStatusJob1-Node2 --> ImageJob1;
ImageStatusJob1-NodeN --> ImageJob1;
ImageJob1 --> ImageList1;
ImageList2 --> ImageJob2;
ImageJob2 --> EraserPodJob2-Node1;
ImageJob2 --> EraserPodJob2-Node2;
ImageJob2 --> EraserPodJob2-NodeN;
EraserPodJob2-Node1 --> ImageStatusJob2-Node1;
EraserPodJob2-Node2 --> ImageStatusJob2-Node2;
EraserPodJob2-NodeN --> ImageStatusJob2-NodeN;
ImageStatusJob2-Node1 --> ImageJob2;
ImageStatusJob2-Node2 --> ImageJob2;
ImageStatusJob2-NodeN --> ImageJob2;
ImageJob2 --> ImageList2;
```

However then we have this extra object that is effectively reporting exactly what the pod already encapsulates since we'd have 1 `imagestatus` per `imagelist` per node.
Which we only end up collecting to throw into `imagelist`'s status anyway.

At this point I also have to wonder if anyone actually needs the detailed status.
On success of a job then the inventory of images on a node will be updated by virtue of that success.
On failure a status is useful for debugging purposes, however we can keep the `imagejob` (and by extension the failed pods for the job) so an admin can look at pod logs and whatever else we end up providing.

To help surface details, the job controller can emit events (for humans) and/or metrics on success or failure.
Such metrics would really only be to help surface errors on the cluster since an image removal error is likely indicative of other problems.
Given that there is one `imagejob` per `imagelist` at a time, the `imagejob` can post the last status in the `imagelist` as it does today (where status would be at most a summary + last run timestamp).

Currently this also means the eraser pod requires cluster access so it can post an `imagestatus` (it also does to get the `imagelist` spec that it is running for... but this can be solved in other ways). Ideally this pod does not require any special privileges except on the node itself to communicate with CRI.

----

TLDR; I think we can get rid of `imagestatus` and the current implementation of `imagestatus` is error prone as it is. Removing `imagestatus` is also a step towards not requiring any cluster privileges for the eraser pod.

### LINKED ISSUE #102: Remove imagestatus CR
`ImageStatus` adds a lot of overhead to running eraser for mostly low-value information.
The intention of this was to have a sort of audit trail as to what happened. It doesn't exactly work that way because details get overwritten and other things related to the implementation, but the intention is to know exactly what happened during a run.

Instead we can change our job deletion strategy with the intention of keeping pods for jobs around longer to ensure pod logs are kept around for a period of time for debugging purposes.
Additionally we should make sure that we log everything an eraser pod does to stdout.

Related #96

cpuguy83:
This replaces imagestatus with a smaller summary of the last run.
Also instead of deleting jobs immediately after completion there are
options to set a delay period that the controller should wait before
deleting.
Detailed information about what was deleted or not is output the the
eraser pod's logs now.

Closes #102 
Closes #96
Closes #60


controllers/imagejob/imagejob_controller.go
```diff
@@ -197,40 +214,40 @@ func (r *Reconciler) handleRunningJob(ctx context.Context, imageJob *eraserv1alp
...
 	if err != nil {
 		return err
 	}
-	if err := r.updateImageListStatus(ctx, nodeResult, imageList); err != nil {
-		return err
-	}
-	return nil
+
+	now := metav1.Now()
+	imageList.Status.Timestamp = &now
```
cpuguy83, line 243: I'm thinking we could put this in a list of statuses from the last... maybe 5 runs + the job name for each run... but maybe a separate PR.

api/v1alpha1/imagejob_types.go
```diff
@@ -53,6 +53,9 @@ type ImageJobStatus struct {
 
 	// job running, successfully completed, or failed
 	Phase JobPhase `json:"phase"`
+
+	// Time to delay deletion until
+	DeleteAfter *metav1.Time `json:"delete_after,omitempty"`
```
sozercan:
```suggestion
	DeleteAfter *metav1.Time `json:"deleteAfter,omitempty"`
```

controllers/imagejob/imagejob_controller.go
```diff
@@ -54,22 +55,33 @@ const (
 
 var log = logf.Log.WithName("controller").WithValues("process", "imagejob-controller")
 
+var (
+	successDelDelaySeconds = flag.Int64("job-cleanup-on-success-delay", 0, "Duration string for how long to delay job deletion for successful runs.")
+	errDelDelaySeconds     = flag.Int64("job-cleanup-on-error-delay", 86400, "Duration string for how long to delay job deletion for errored runs.")
+	successeRatio          = flag.Float64("job-success-ratio", 1.0, "Ratio of successful runs to consider a job successful")
```
sozercan:
```suggestion
	successRatio          = flag.Float64("job-success-ratio", 1.0, "Ratio of successful runs to consider a job successful")
```

controllers/imagejob/imagejob_controller.go
```diff
@@ -54,22 +55,33 @@ const (
 
 var log = logf.Log.WithName("controller").WithValues("process", "imagejob-controller")
 
+var (
+	successDelDelaySeconds = flag.Int64("job-cleanup-on-success-delay", 0, "Duration string for how long to delay job deletion for successful runs.")
+	errDelDelaySeconds     = flag.Int64("job-cleanup-on-error-delay", 86400, "Duration string for how long to delay job deletion for errored runs.")
```
sozercan: we should mention these are in seconds in description

controllers/imagejob/imagejob_controller.go
sozercan: Can we describe this more? For 1.0, if there's a single error, that means it's not a sucess?

controllers/imagejob/imagejob_controller.go
```diff
@@ -54,22 +55,33 @@ const (
 
 var log = logf.Log.WithName("controller").WithValues("process", "imagejob-controller")
 
+var (
+	successDelDelaySeconds = flag.Int64("job-cleanup-on-success-delay", 0, "Duration string for how long to delay job deletion for successful runs.")
```
sozercan: what's 0 here? delete immediately or never delete?

controllers/imagejob/imagejob_controller.go
```diff
@@ -306,12 +332,16 @@ func (r *Reconciler) handleNewJob(ctx context.Context, imageJob *eraserv1alpha1.
...
+		if !fitness {
+			log.Info("Eraser pod does not fit on node, skipping")
+			continue
+		}
+
+		err = r.Create(ctx, pod)
+		if err != nil {
+			return err
 		}
+		log.Info("Started earser pod on node", "images", ls.Spec.Images)
```
sozercan:
```suggestion
		log.Info("Started eraser pod on node", "images", ls.Spec.Images)
```

controllers/imagejob/imagejob_controller.go
```diff
@@ -197,40 +214,40 @@ func (r *Reconciler) handleRunningJob(ctx context.Context, imageJob *eraserv1alp
...
+		DeleteAfter: after(time.Now(), *successDelDelaySeconds),
 	}
 
-	// transfer results from imageStatus objects to imageList
-	statusList := &eraserv1alpha1.ImageStatusList{}
-	err = r.List(ctx, statusList)
-	if err != nil {
-		return err
+	if float64(success/imageJob.Status.Desired) < *successeRatio {
+		log.Info("Marking job as failed", "success-ratio", *successeRatio, "actual ratio", success/imageJob.Status.Desired)
```
sozercan:
```suggestion
		log.Info("Marking job as failed", "success ratio", *successeRatio, "actual ratio", success/imageJob.Status.Desired)
```

cpuguy83:
Fixed all the things.

sozercan (approved):
LGTM

PR #111, opened 2022-02-26
Configure eraser pod log

cpuguy83:
This was not being configured and log entries were being lost.
This moves all the setup to a common package to make sure log setup is
the same for everything.

### DIFF
--- a/controllers/imagejob/imagejob_controller.go
+++ b/controllers/imagejob/imagejob_controller.go
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	eraserv1alpha1 "github.com/Azure/eraser/api/v1alpha1"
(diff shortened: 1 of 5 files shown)


controllers/imagejob/imagejob_controller.go
```diff
@@ -291,7 +292,7 @@ func (r *Reconciler) handleNewJob(ctx context.Context, imageJob *eraserv1alpha1.
 
 		givenImage := imageJob.Spec.JobTemplate.Spec.Containers[0]
 		image := corev1.Container{
-			Args:            append(givenImage.Args, "--runtime="+runtimeName),
+			Args:            append(givenImage.Args, []string{"--runtime=" + runtimeName, "--log-level=" + logger.GetLevel()}...),
```
sozercan: can we create a var for this and add here

sozercan (approved):
minor nit, otherwise LGTM

PR #115, opened 2022-02-28
Do not require cluster access for eraser pod

### LINKED ISSUE #114: Remove need for kube API access in eraser pod
The eraser pod used to do more things like post status updates.
Now the only thing it uses API access for is to get the image list.

We should be able to remove this requirement.

cpuguy83:
Makes eraser read the imagelist from a configmap.
This does a few things:

1. No longer need cluster access
2. Provides an immutable list of images that every eraser pod uses for a
   given job.
3. Reduces requests on the cluster

One downside of this is we store the list multiple times (in the
imagelist and a configmap for each job run).

Closes #114


sozercan (approved):
LGTM

cpuguy83:
Rebased

PR #118, opened 2022-03-02
Add eraser unit tests

cpuguy83:
Previous test cases were just explicitly for the test client, but the
test client was never actually used to test anything except itself.

This adds a bunch of test cases for `removeImages`.


sozercan (approved):
LGTM

PR #133, opened 2022-04-12
Use fixed go version to install setup-envtest

pmengelbert:
`go get` and `go install` have different behaviors and different
interpretations of environment variables such as GOBIN. Since we are
already using a go image with a fixed version to fetch `controller-gen`,
we can use the same container to fetch the `setup-envtest` binary.

This avoids the problem of `go get` behaving the wrong way if you have
the "wrong" version of go installed.

Since the tests are run locally (in a non-container environment), mount
the $PROJECT_DIR/bin directory to obtain the `setup-envtest` binary for
the local environment.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,8 +158,10 @@ promote-staging-manifest:
 
 ENVTEST = $(shell pwd)/bin/setup-envtest
 .PHONY: envtest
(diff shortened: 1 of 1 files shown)


sozercan (approved):
Thank you! LGTM

PR #135, opened 2022-04-12
Imagelist reconcile on update

### LINKED ISSUE #126: imagelist only gets reconciled on create but not update
we should reconcile on update too

cc @cpuguy83 

**Environment:**

- Eraser version:

ashnamehrotra:
Imagelist will reconcile on create and update, and ignore reconcile on status changes by checking generation changed predicate as an event filter.

Fixes #126


controllers/imagelist/imagelist_controller.go
```diff
@@ -139,7 +137,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 	err = c.Watch(
 		&source.Kind{Type: &eraserv1alpha1.ImageList{}},
-		&handler.EnqueueRequestForObject{})
+		&handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{})
```
pmengelbert, line 143: So that I understand this, `predicate.GenerationChangedPredicate()` will cause the `Reconcile()` function to be run only if the ImageList's `spec` field changes, but not its status. Is that right?

pmengelbert (changes requested):
It would be good to see some tests (unit and e2e) to accompany these code changes. It's a small change, but it fixes a bug and tests that check for action on update will prevent us from accidentally regressing in the future.

controllers/imagelist/imagelist_controller.go
ashnamehrotra: Yup that's right, status updates won't change ObjectMeta.Generation but spec changes will

ashnamehrotra:
The test first deletes nginx, and then updates imagelist to prune all remaining images. I also ran into a timeout issue when updating imagelist to delete the specific redis image instead of prune all remaining, which I think is related to the issue @pmengelbert was working on with the intermittent timeouts.

test/e2e/test-data/eraser_v1alpha1_imagelist_updated.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: eraser.sh/v1alpha1
+kind: ImageList
+metadata:
+  name: imagelist
+spec:
+  images:
+    - "*"
```
cpuguy83: Nit: new line

cpuguy83 (commented):
Can you think about how we might test that a status update is not triggering a reconcile?

cpuguy83:
Looks like tests are timing out when checking for image removal, even before status update.
Building locally to see if I can get more details.

cpuguy83:
It looks like the predicate handler is preventing the requeue logic in `handleCompletedJob` from working. 
It's kind of strange that we are hitting the logic since this is only set in the future when there the success ratio is not being met... but the success ratio *is* met.

cpuguy83:
In any case, it seems this handler breaks cases where we need to delay deletion of the imagejob.

controllers/imagelist/imagelist_controller.go
```diff
@@ -91,39 +92,36 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	// Check to make sure reconcile isn't from updating ImageStatus
-	if imageList.Status.Timestamp == nil {
-		// If there is a change in ImageList, start ImageJob to trigger removal
-		job := &eraserv1alpha1.ImageJob{
-			ObjectMeta: metav1.ObjectMeta{
-				GenerateName: "imagejob-",
-				OwnerReferences: []metav1.OwnerReference{
```
cpuguy83, line 100: Here's a problem. This isn't being set in the new version

cpuguy83 (approved):
LGTM

cpuguy83:
@pmengelbert Can you give this a look?

pmengelbert (changes requested):
Needs a rebase on main, but otherwise looks good to me

pmengelbert (approved):
👍

Issue #134, opened 2022-04-12
e2e test has intermittent failures


pmengelbert:
**What steps did you take and what happened:**
```bash
$ make docker-build-manager docker-build-eraser OUTPUT_TYPE="type=image"
$ for i in {0..5}; do make e2e-test; done
```

What happened is that the tests failed intermittently.

**What did you expect to happen:**
I expected the tests to pass every time, or to fail every time.

**Anything else you would like to add:**

**Environment:**

- Eraser version: v0.1.0
- Kubernetes version: (use `kubectl version`): 
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:58:47Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.0", GitCommit:"ab69524f795c42094a6630298ff53f3c3ebab7f4", GitTreeState:"clean", BuildDate:"2021-12-08T00:29:07Z", GoVersion:"go1.17.3", Compiler:"gc", Platform:"linux/amd64"}
```

pmengelbert:
The failure I experienced was `eraser_test.go:254: timeout waiting for images to be cleaned`.

It could be that the process just needs a longer timeout. It could be that something else is going wrong which is *causing* the timeout. More detailed logging about the state of the program at timeout will help.

pmengelbert:
After I ran the tests, I got the following:
```bash
$ kubectl logs -n eraser-system eraser-eraser-e2e-test-worker2 |jq
{
  "level": "info",
  "ts": 1650037176.5409193,
  "logger": "eraser",
  "msg": "Image is not on node",
  "image": "sha256:2834dc507516af02784808c5f48b7cbe38b8ed5d0f4837f16e78d00deb7e7767"
}
{
  "level": "info",
  "ts": 1650037176.666359,
  "logger": "eraser",
  "msg": "Removed",
  "image": "docker.io/library/nginx:latest"
}
{
  "level": "info",
  "ts": 1650037176.6663945,
  "logger": "eraser",
  "msg": "Image is not on node",
  "image": "nginx"
}

$ docker exec eraser-e2e-test-worker2 ctr -n k8s.io images list  | grep nginx
[...]

PR #141, opened 2022-04-15
Add node selector to image job pod spec

### LINKED ISSUE #49: add nodeSelector for linux

### LINKED ISSUE #130: An image may not be deleted if it is an alias for another image

```bash
$ docker pull docker.io/library/nginx:latest
$ docker tag docker.io/library/nginx:latest registry.test/nginx:one
$ docker tag docker.io/library/nginx:latest registry.test/nginx:two
$ docker push registry.test/nginx:one
$ docker push registry.test/nginx:two
$ cat <<'EOF' | kubectl apply -f -
---
apiVersion: v1
kind: Pod
metadata:
  name: nginx1
spec:
  containers:
    - image: registry.test/nginx:one
      name: nginxone
  nodeName: mynode1
---
apiVersion: v1
kind: Pod
metadata:
  name: nginx2
spec:
  containers:
    - image: registry.test/nginx:two
      name: nginxtwo
  nodeName: mynode1
EOF

$ kubectl delete pod nginx1 nginx2
$ cat <<'EOF' | kubectl apply -f -
---
apiVersion: eraser.sh/v1alpha1
kind: ImageList
metadata:
  name: imagelist
spec:
  images:
    - registry.test/nginx:two
EOF

$ kubectl logs -n eraser-system eraser-mynode1 | jq
...
{
  "level": "info",
  "ts": 1649354133.998637,
  "logger": "eraser",
  "msg": "Image is not on node",
  "image": "registry.test/nginx:two"
}
...
```

In this case, registry.test/nginx:one and registry.test/nginx:two are aliases
for each other, and supplying the following to the imagelist_controller should
result in the underlying image being deleted:

```
spec:
  images:
    - registry.test/nginx:two
```

This is a result of the following code (`pkg/eraser/eraser.go:215`):

```go
	nonRunningNames := make(map[string]struct{}, len(allImages)-len(runningImages))
	for key := range nonRunningImages {
		if idMap[key] != nil && len(idMap[key]) > 0 {
			nonRunningNames[idMap[key][0]] = struct{}{}
		}
	}
```

Only the first image name is added to the set of images that aren't running. In
the program, `idMap[key]` looks like:

```json
  [
    "pengelbert.azurecr.io/nginx:one",
    "pengelbert.azurecr.io/nginx:two"
  ]
```

According to the above code, only `nginx:one` is added to the set of
non-running images. As a result, no match will be found between a) the provided
imagelist and b) the set of non-running images.

**Environment:**

- Eraser version: v0.1.0

```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:58:47Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.9", GitCommit:"56709e92afa973c26fad3d4a44723fefa51481b7", GitTreeState:"clean", BuildDate:"2022-03-10T07:59:33Z", GoVersion:"go1.16.12", Compiler:"gc", Platform:"linux/amd64"}
WARNING: version difference between client (1.23) and server (1.21) exceeds the supported minor version skew of +/-1
```

pmengelbert:
The driving factor of this PR is that we want to avoid scheduling eraser pods on Windows nodes. This is because we do not yet have scanning tools for Windows, and the eraser pods are built to run on Linux. We may in the future deploy these pods to Windows nodes, in which case changes to this PR will need to be made.

In order to make the scheduling system a little more flexible (and make testing easier), we will also skip nodes as requested by the user. The mechanics of this are as follows:
* Each --skip-nodes-selector flag supplies a single selector. Each
  selector may match one or more label.
* If a node matches **one** of the selectors (*not* labels), the node
  will be skipped.
* There are two defaults that cannot be overridden. Those two defaults
  are `kubernetes.io/os=windows` and `eraser.sh/cleaup.skip` (it doesn't
  matter what the value is for the latter, if the key is present, we
  will skip the node).

If we support Windows nodes in the future, the `kubernetes.i/os=windows` will be removed from the defaults; or, at the very least, we will allow the user to override that flag.

Comments on this PR are welcome and encouraged. The tests should probably be refactored and cleaned up a bit.

Resolves #49 


test/e2e/eraser_test.go
```diff
@@ -86,6 +91,7 @@ func TestRemoveImagesFromAllNodes(t *testing.T) {
 			}
 
 			// deploy imageJob config
+			time.Sleep(45 * time.Second)
```
sozercan: why are we sleeping here? can we watch instead of sleep?

Makefile
```diff
@@ -98,7 +100,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 # Run e2e tests
 .PHONY: e2e-test
 e2e-test:
-	IMAGE=${ERASER_IMG} MANAGER_IMAGE=${MANAGER_IMG} NODE_VERSION=kindest/node:v${KUBERNETES_VERSION} go test -count=1 -tags=e2e -v ./test/e2e
+	CGO_ENABLED=0 IMAGE=${ERASER_IMG} MANAGER_IMAGE=${MANAGER_IMG} NODE_VERSION=kindest/node:v${KUBERNETES_VERSION} go test -count=$(TEST_COUNT) -timeout 86400s -tags=e2e -v ./test/e2e
```
sozercan: do we need such a long timeout?

api/v1alpha1/imagelist_types.go
```diff
@@ -31,6 +31,8 @@ type ImageListStatus struct {
 	Success int64 `json:"success"`
 	// Number of nodes that failed to run the job
 	Failed int64 `json:"failed"`
+	// Number of nodes that were skipped, e.g. because they are a non-linux node
```
sozercan: let's keep it generic (say "due to skip selector" or something), we can specific non-Linux in docs

Makefile
pmengelbert: No, that was set to address some local issue. 

- [x] I'll reduce it

api/v1alpha1/imagelist_types.go
pmengelbert: 👍 

- [x] Fix comment

controllers/imagejob/imagejob_controller.go
```diff
@@ -276,12 +286,35 @@ func (r *Reconciler) handleNewJob(ctx context.Context, imageJob *eraserv1alpha1.
...
 
 		nodeName := nodes.Items[i].Name
+		for _, skipNodesSelector := range skipNodesSelectors {
+			skipLabels, err := labels.Parse(skipNodesSelector)
+			if err != nil {
+				return err
+			}
+
+			log.Info("skipLabels", "skipLabels", skipLabels)
+			log.Info("nodeLabels", "nodeLabels", nodes.Items[i].ObjectMeta.Labels)
```
sozercan: these seem too verbose for Info, should we remove or move debug?

test/e2e/eraser_test.go
pmengelbert: I just made a comment on #134 to this effect. I am sleeping because the `wait.For` doesn't actually result in the right state for deletion. The images still appear to eraser to be `running`. We will have to ditch the `wait.For` calls in favor of something else.

I only used `time.Sleep` to verify that the issue was a race condition of some kind. How would you recommend checking that the pods are absolutely done?

controllers/imagejob/imagejob_controller.go
pmengelbert: Preferable to move it to debug, as it's useful for troubleshooting.

- [x] I'll look into how to manage verbose logging.

pmengelbert: Will now only appear when --log-level=debug is set on the pod

test/e2e/eraser_test.go
pmengelbert: - [x] Determine a better way to watch for pod & deployment deletion before deploying eraser in tests.

pmengelbert:
Rebased on main

pkg/eraser/eraser.go
```diff
@@ -195,11 +195,16 @@ func removeImages(c Client, targetImages []string) error {
...
-	// map for non-running imags by name
-	nonRunningNames := make(map[string]struct{}, len(allImages)-len(runningImages))
-	for key := range nonRunningImages {
-		if idMap[key] != nil && len(idMap[key]) > 0 {
-			nonRunningNames[idMap[key][0]] = struct{}{}
+	log.V(1).Info("List of non-running images by digest", "nonRunningImages", nonRunningImages)
+	log.V(1).Info("Map of digest to image name(s)", "idMap", idMap)
+
+	// mapping of non-running image names to digests
+	nonRunningNames := make(map[string]string, len(allImages)-len(runningImages))
```
cpuguy83: Can you add a unit test for what's being fixed here?

controllers/imagejob/imagejob_controller.go
```diff
@@ -63,8 +63,13 @@ var (
 	successDelDelaySeconds = flag.Int64("job-cleanup-on-success-delay", 0, "Seconds to delay job deletion after successful runs. 0 means no delay")
 	errDelDelaySeconds     = flag.Int64("job-cleanup-on-error-delay", 86400, "Seconds to delay job deletion after errored runs. 0 means no delay")
 	successRatio           = flag.Float64("job-success-ratio", 1.0, "Ratio of successful/total runs to consider a job successful. 1.0 means all runs must succeed.")
+	skipNodesSelectors     = nodeSkipSelectors([]string{"kubernetes.io/os=windows", "eraser.sh/cleanup.skip"})
 )
 
+func init() {
```
cpuguy83, line 69: How come this is in a `func init()`?

controllers/imagejob/imagejob_controller.go
cpuguy83: Shouldn't these be set to `log.Debug`?

pkg/eraser/eraser.go
```diff
@@ -195,11 +195,16 @@ func removeImages(c Client, targetImages []string) error {
 		}
 	}
 
-	// map for non-running imags by name
-	nonRunningNames := make(map[string]struct{}, len(allImages)-len(runningImages))
-	for key := range nonRunningImages {
-		if idMap[key] != nil && len(idMap[key]) > 0 {
-			nonRunningNames[idMap[key][0]] = struct{}{}
+	log.V(1).Info("List of non-running images by digest", "nonRunningImages", nonRunningImages)
```
cpuguy83: Probably should be debug only.

cpuguy83 (dismissed):

controllers/imagejob/imagejob_controller.go
pmengelbert: I suppose it should be in the `var` block, with the other flags, in `main.go`. I'll move it.

cpuguy83: Sorry, I misread this...
flag.Var is fine here... my eyes were just seeing it as us setting a flag value here.

controllers/imagejob/imagejob_controller.go
pmengelbert: The `Logger` struct doesn't have a `Debug()` function, but we can put them at a higher level with `log.V(1).Info(...)`, which hides them from pod logs unless `--log-level=debug` flag is provided. I'll do that unless there's a better way.

controllers/imagejob/imagejob_controller.go
pmengelbert: See above, log.V(1).Info is a way to get debug-only logging here, as it raises the verbosity level by one. Since there's no .Debug() method, I wasn't sure how to go about that here.

pkg/eraser/eraser.go
pmengelbert: Will do, thanks

- [ ] Unit test

pkg/eraser/eraser.go
pmengelbert: See above, `log.V(1).Info` is a way to get debug-only logging here, as it raises the verbosity level by one. Since there's no `.Debug()` method, I wasn't sure how to go about that here.

pkg/eraser/eraser.go
pmengelbert: After taking a look, I don't think a Unit Test is going to work here, since the `testClient` isn't aware of digests or Aliases. Either we make the `testClient` aware and have this as a unit test, or we write an e2e test for this. Either way, I'm going to separate this into its own PR, since it will be useful to Ashna for her PR as well.

pmengelbert:
Rebased onto #143 , and that should be merged first

pkg/eraser/eraser.go
pmengelbert: Separated into another PR, which will be put up when the test is done, so marking as resolved.

controllers/imagejob/imagejob_controller.go
```diff
@@ -386,6 +419,13 @@ func (r *Reconciler) handleNewJob(ctx context.Context, imageJob *eraserv1alpha1.
 		}
 		log.Info("Started eraser pod on node", "images", imgList.Spec.Images)
 	}
+
+	imageJob.Status.Skipped = skipped
+	if err := r.updateJobStatus(ctx, imageJob); err != nil {
+		log.Error(err, "failure updating job status")
```
cpuguy83: This should either return an error or log the error, but not both.

pmengelbert: OK, we'll bubble the error up.

test/e2e/eraser_test.go
```diff
@@ -583,8 +588,172 @@ func TestRemoveImagesFromAllNodes(t *testing.T) {
...
+				t.Error("Failed to deploy image list config", err)
+			}
+
+			ctxT, cancel := context.WithTimeout(ctx, time.Minute)
+			defer cancel()
+
+			// ensure images are removed from all nodes except the one we are skipping. remove the node we are skipping from the list of nodes.
+
+			checkImageRemoved(ctxT, t, clusterNodes, nginx)
+
```
cpuguy83, line 700: Can we check that the image is not removed from the skipped node?

pmengelbert: Done

cpuguy83 (approved):
LGTM

PR #155, opened 2022-04-22
Update to support multi-arch builds

### LINKED ISSUE #151: add support for arm64
eraser should support arm64 arch

jeremyrickard:
Signed-off-by: Jeremy Rickard <jrickard@microsoft.com>

This PR updates the Dockerfile/Makefile to support building arm64 in addition to amd64. 

Highlights:

* Update Dockerfile  to use the TARGETOS and TARGETARCH arguments passed by build kit in order to properly compile for the selected platform (i.e. linux/arm64 or linux/amd64). 
* Updates the FROM statements to use --platform and moves the base images to arguments as well. 
* Updates Makefile to move platform to a variable to allow users to override when using the make targets

This doesn't update the Makefile to produce linux/arm64 containers as part of a release process but might be good to do that in a follow up? 

Fixes #151 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,11 @@
 # syntax=mcr.microsoft.com/oss/moby/dockerfile:1.3.1
-ARG BUILDPLATFORM="linux/amd64"
 ARG BUILDERIMAGE="golang:1.17"
 
 ARG ERASERBASEIMAGE="gcr.io/distroless/static:latest"
 ARG MANAGERBASEIMAGE="gcr.io/distroless/static:nonroot"
 
-ARG TARGETOS=linux
-ARG TARGETARCH=amd64
+ARG TARGETOS
+ARG TARGETARCH
 
 # Build the manager binary
 FROM --platform=$BUILDPLATFORM $BUILDERIMAGE AS builder
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,15 @@
 # syntax=mcr.microsoft.com/oss/moby/dockerfile:1.3.1
+ARG BUILDPLATFORM="linux/amd64"
+ARG BUILDERIMAGE="golang:1.17"
+
+ARG ERASERBASEIMAGE="gcr.io/distroless/static:latest"
+ARG MANAGERBASEIMAGE="gcr.io/distroless/static:nonroot"
+
+ARG TARGETOS=linux
+ARG TARGETARCH=amd64
 
 # Build the manager binary
-FROM golang:1.17 AS builder
+FROM --platform=$BUILDPLATFORM $BUILDERIMAGE AS builder
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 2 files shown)


pmengelbert (commented):
LGTM

Dockerfile
```diff
@@ -1,7 +1,15 @@
 # syntax=mcr.microsoft.com/oss/moby/dockerfile:1.3.1
+ARG BUILDPLATFORM="linux/amd64"
```
cpuguy83: This var is set by buildkit, no need to set it here.

Dockerfile
```diff
@@ -1,7 +1,15 @@
...
+ARG BUILDERIMAGE="golang:1.17"
+
+ARG ERASERBASEIMAGE="gcr.io/distroless/static:latest"
+ARG MANAGERBASEIMAGE="gcr.io/distroless/static:nonroot"
+
+ARG TARGETOS=linux
```
cpuguy83: Same as above, this is set by buildkit.

Makefile
```diff
@@ -9,6 +9,8 @@ KUBERNETES_VERSION ?= 1.23.0
 ENVTEST_K8S_VERSION ?= 1.23
 GOLANGCI_LINT_VERSION := 1.43.0
 
+PLATFORMS ?= linux/amd64
```
cpuguy83: Should be singular

cpuguy83 (changes requested):

Makefile
jeremyrickard: 👍

Dockerfile
```diff
@@ -1,7 +1,14 @@
 # syntax=mcr.microsoft.com/oss/moby/dockerfile:1.3.1
+ARG BUILDERIMAGE="golang:1.17"
+
+ARG ERASERBASEIMAGE="gcr.io/distroless/static:latest"
+ARG MANAGERBASEIMAGE="gcr.io/distroless/static:nonroot"
+
+ARG TARGETOS
```
cpuguy83: These are essentially unused.
They need to be set within the `FROM` that they are going to be consumed in.

cpuguy83 (changes requested):

Dockerfile
jeremyrickard: omg, thanks for catching that!

Makefile
```diff
@@ -9,6 +9,8 @@ KUBERNETES_VERSION ?= 1.23.0
 ENVTEST_K8S_VERSION ?= 1.23
 GOLANGCI_LINT_VERSION := 1.43.0
 
+PLATFORM ?= linux/amd64
```
cpuguy83: One last nit, default value for this can be `linux`.
This way if you are building on arm64 it doesn't try to build amd64.

jeremyrickard: Oh thanks, that's great to know!

cpuguy83 (approved):
LGTM

PR #161, opened 2022-04-28
Only delay deltion when until is sufficiently long

cpuguy83:
Found a case where requeue wasn't actually occurring because
`RequeueAfter` was being set to 0 because we recalculated the delay
after we checked it (and calculated it slightly differently).
This casued deletion to never happen since `RequeueAfter` must be > 0 to
actually trigger a requeue..


pmengelbert (approved):
Nice catch, LGTM.

PR #173, opened 2022-05-12
Scanner

pmengelbert:
- Add new `scanner` binary that uses trivy to scan for vulnerabilities in images.
- Enable scanner to read from Collector CRs, and write to Collector CR status.
- Provide make targets to build scanner images.

- In the interest of moving things along, this PR uses the REST client and the manager's `serviceaccount`. I opened #177 to track this


pkg/scanner/scanner.go
```diff
@@ -0,0 +1,345 @@
+package main
```
sozercan: should this go under something like /pkg/scanners/trivy/trivy.go?

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	app := applier.NewApplier(filesystemCache)
+	det := ospkg.Detector{}
+	dopts := fanalTypes.DockerOption{}
+	scan := local.NewScanner(app, det)
+
+	sopts := trivyTypes.ScanOptions{
+		VulnType:            []string{"os", "library"},
+		SecurityChecks:      []string{"vuln", "secret"},
```
sozercan: vulnType and securityChecks should be configurable with flags

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,345 @@
...
+		fmt.Println(imageRef)
+	}
+}
+
+func parseSeverity(sevString string) error {
+	sevs := strings.Split(sevString, ",")
+	for _, sev := range sevs {
+		_, ok := severityMap[sev]
+		if !ok {
+			return fmt.Errorf("severity '%s' should be one of of [CRITICAL, HIGH, MEDIUM, LOW, UNKNOWN]", sev)
```
sozercan:
```suggestion
			return fmt.Errorf("severity '%s' should be one of [CRITICAL, HIGH, MEDIUM, LOW, UNKNOWN]", sev)
```

Dockerfile
```diff
@@ -30,6 +31,16 @@ RUN \
     --mount=type=cache,target=/go/pkg/mod \
     GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o out/manager main.go
 
+FROM builder AS scanner-build
+
+ARG TARGETOS
```
sozercan, line 39: can we define ARG at the top? do these need to be duplicated?

Dockerfile
```diff
@@ -3,6 +3,7 @@ ARG BUILDERIMAGE="golang:1.17"
 
 ARG ERASERBASEIMAGE="gcr.io/distroless/static:latest"
 ARG MANAGERBASEIMAGE="gcr.io/distroless/static:nonroot"
+ARG SCANNERBASEIMAGE="gcr.io/distroless/static:nonroot"
```
sozercan: instead of generic "scanner", let's say "TRIVY_SCANNER" or something

sozercan: applies to everywhere we say "scanner"

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,345 @@
...
+	apiPath         = "apis/eraser.sh/v1alpha1"
+	resourceName    = "imagecollectors"
+	subResourceName = "status"
+)
+
+var (
+	collectorCRName = flag.String("collector-cr-name", "collector-cr", "name of the collector cr to read from and write to")
+	cacheDir        = flag.String("cache-dir", "/var/lib/trivy", "path to the cache dir")
+	severity        = flag.String("severity", "CRITICAL,HIGH,MEDIUM,LOW,UNKNOWN", "list of severity levels to report")
+	ignoreUnfixed   = flag.Bool("ignore-unfixed", false, "report only fixed vulnerabilities")
```
sozercan: should we default to true?

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,345 @@
...
+
+	apiPath         = "apis/eraser.sh/v1alpha1"
+	resourceName    = "imagecollectors"
+	subResourceName = "status"
+)
+
+var (
+	collectorCRName = flag.String("collector-cr-name", "collector-cr", "name of the collector cr to read from and write to")
+	cacheDir        = flag.String("cache-dir", "/var/lib/trivy", "path to the cache dir")
+	severity        = flag.String("severity", "CRITICAL,HIGH,MEDIUM,LOW,UNKNOWN", "list of severity levels to report")
```
sozercan: should we default to CRITICAL?

sozercan: thoughts @miwithro?

pkg/scanner/scanner.go
pmengelbert: 👍

pkg/scanner/scanner.go
pmengelbert: Trivy default is `CRITICAL,HIGH`, I'm fine with either that or `CRITICAL`

miwithro: @sozercan I think if we hit MEDIUM, HIGH, CRITICAL we will be able to deliver on customer expectations.

pkg/scanners/trivy/trivy.go
```diff
@@ -0,0 +1,399 @@
+package main
```
sozercan, line 1: as we discussed before, let's add unit tests

we can add e2e once we have the collector logic

pkg/scanners/trivy/trivy.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	go func() {
+		wg.Wait()
+		close(imgChan)
+	}()
+
+	vulnerableImages := make([]eraserv1alpha1.Image, 0, len(scanList))
+	for imageRef := range imgChan {
+		image := eraserv1alpha1.Image{Digest: "abc123", Name: imageRef}
```
sozercan: placeholder?

pkg/scanners/trivy/trivy.go
```diff
@@ -0,0 +1,397 @@
...
+			wg.Done()
+		}(imageName)
+	}
+
+	go func() {
+		wg.Wait()
+		close(imgChan)
+	}()
+
+	vulnerableImages := make([]eraserv1alpha1.Image, 0, len(scanList))
```
sozercan: should we separate out the reporting back logic so it can be reused later?

pkg/scanners/trivy/trivy.go
```diff
@@ -0,0 +1,397 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func downloadDB(cacheDir string) error {
+	client := dlDb.NewClient(cacheDir, true)
+	ctx := context.Background()
+	needsUpdate, err := client.NeedsUpdate("dev", false)
```
sozercan: what's `dev` here?

pmengelbert: It refers to the trivy version. This made its way in from looking at Trivy's state when it was running. I've corrected this now.

pkg/scanners/trivy/trivy.go
pmengelbert: @sozercan Would you mind elaborating? I'm not sure I follow what you mean

pkg/scanners/trivy/trivy.go
```diff
@@ -0,0 +1,435 @@
...
+	}
+
+	// Will be modified by parseCommaSeparatedOptions()  to reflect the `vuln-type` CLI flag
+	// These are the only recognized vulnerability types and the keys of this map should never be modified.
+	vulnTypeMap = map[string]bool{
+		vulnTypeOs:      false,
+		vulnTypeLibrary: false,
+	}
+
+	log = logf.Log.WithName("eraser")
```
sozercan: do we want to say "scanner" here? and perhaps `WithValues("provider", "trivy")`?

pkg/scanners/trivy/trivy.go
sozercan: each provider processes collector results and outputs a list, right? say, if we have a different scanner tomorrow, can we share logic for the output of a list?

pkg/scanners/trivy/trivy.go
```diff
@@ -0,0 +1,435 @@
...
+		go func(img eraserv1alpha1.Image) {
+			imageRef := img.Name
+			if imageRef == "" {
+				log.Info("found image with no name", "img", img)
+				imgChan <- imageReport{Image: img, err: fmt.Errorf("no name")}
+				wg.Done()
+				return
+			}
+
+			fmt.Printf("scanning: %s\n", imageRef)
```
sozercan: should we use log here?

pkg/scanners/trivy/trivy.go
```diff
@@ -0,0 +1,435 @@
...
+	vulnTypeLibrary = "library"
+
+	securityCheckVuln   = "vuln"
+	securityCheckConfig = "config"
+	securityCheckSecret = "secret"
+
+	apiPath         = "apis/eraser.sh/v1alpha1"
+	resourceName    = "imagecollectors"
+	subResourceName = "status"
+	trivyVersion    = "v0.24.4"
```
sozercan: i am not familiar with this, does this need to match the go.mod version of trivy?

pkg/scanners/trivy/trivy.go
```diff
@@ -0,0 +1,435 @@
...
+		logMsg := "image has vulnerabilities"
+
+		if img.err != nil {
+			log.V(1).Error(img.err, "error scanning image") // debug only
+			img.ScanSucceeded = boolPtr(false)
+			logMsg = "unable to scan image"
+		}
+
+		log.Info(logMsg, "img", img)
+		vulnerableImages = append(vulnerableImages, img.Image)
```
sozercan, line 294: are we adding images that failed to scan to the vulnerable images list?

pmengelbert: @sozercan We are adding them to the `Results` section of `ImageCollector.Status`, with `ScanSucceeded` set to `false`

pkg/scanners/trivy/trivy.go
pmengelbert: No, not necessarily, but it probably should. Could we set this using LDFlags based on the version in `go.mod`?

sozercan: Looks like trivy gets this info from appVersion, can we use that or is that part of cli only? https://github.com/aquasecurity/trivy/blob/1b66b77f691f2249a63c284056afcad704c0e15f/pkg/commands/option/global.go#L35

pmengelbert: @sozercan I think it's CLI only. I am setting this using ldflags in the makefile:
```
TRIVY_VERSION ?= $(shell go list -f '{{ .Version }}' -m github.com/aquasecurity/trivy)
...
TRIVY_SCANNER_LDFLAGS ?= $(LDFLAGS)-X 'main.trivyVersion=$(TRIVY_VERSION)'
```

pmengelbert: @sozercan This is only used for error output. Trivy's [process](https://github.com/aquasecurity/trivy/blob/06659f1509a19f601d7f7ff0cfe221f87c466566/pkg/db/db.go#L79) for fetching db looks like this:
- Look at the db file on disk, if the db file doesn't exist yet, then this is the first run and it will automatically fetch the latest version of the db, using an OCI client, from `ghcr`. It uses major version tags to fetch the latest version
- the `trivy` library depends on the `trivy-db` package, and that package has a constant `db.Version` ([here](https://github.com/aquasecurity/trivy-db/blob/39d8a6798e072884afa497c4939036dbafd3b2e4/pkg/db/db.go#L20)) set to the major version of the db
- Internally, if the `db.Version` is compared against the `SchemaVersion` found in the db on the filesystem. If `db.Version` is less than `SchemaVersion`, it means we are using an older client with a newer db, and `trivyVersion` will be propagated into the error output.
- If, `db.Version` is >= `SchemaVersion`, then `db.Version` (the version determined by the `trivy-db` version, which is determined by the `trivy` version) will be used as the major version tag for the OCI fetch.
- Finally, timestamps are compared to determine whether or not the db needs to be updated or not.

sozercan (approved):
Thank you! LGTM

PR #171, opened 2022-05-09
Imagecollector crd

ashnamehrotra:
**What this PR does / why we need it**:
Defines ImageCollector CR.

**Which issue(s) this PR fixes** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #165 

**Special notes for your reviewer**:

PR #175, opened 2022-05-13
Collector

### LINKED ISSUE #69: add collector
ListImages() in Container Runtime Interface (CRI) to get a list of images present in a node, then write to a ImageCollectorResult CR containing a list of results from each node.

ashnamehrotra:
Code for collector pods that ImageCollectorController will deploy through ImageJob. 
Uses ListImages() from CRI and creates ImageCollectorCR with list of all images on the node.

Fixes #69 


pkg/collector/collector.go
```diff
@@ -0,0 +1,237 @@
...
+type client struct {
+	images  pb.ImageServiceClient
+	runtime pb.RuntimeServiceClient
+}
+
+type Client interface {
+	listImages(context.Context) ([]*pb.Image, error)
+}
+
+func (c *client) listImages(ctx context.Context) (list []*pb.Image, err error) {
```
sozercan, line 41: looks like several of these functions are duplicated from eraser, can we refactor to share?

.gitignore
```diff
@@ -23,3 +23,6 @@ testbin/*
 *.swp
 *.swo
 *~
+
+PROJECT
```
sozercan: why are we adding `PROJECT` to ignore?

pkg/collector/collector.go
```diff
@@ -0,0 +1,237 @@
...
+		return err
+	}
+
+	_, err = clientset.RESTClient().Post().
+		AbsPath(apiPath).
+		Resource("imagecollectors").
+		Body(body).DoRaw(ctx)
+
+	if err != nil {
+		log.Error(err, "ERROR: Could not create imagecollector", imageCollector.Name, imageCollector.APIVersion)
```
sozercan:
```suggestion
		log.Error(err, "could not create imagecollector", imageCollector.Name, imageCollector.APIVersion)
```

no need to say error

pkg/collector/collector.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	imageCollector := eraserv1alpha1.ImageCollector{
+		TypeMeta: v1.TypeMeta{
+			APIVersion: "eraser.sh/v1alpha1",
+			Kind:       "ImageCollector",
+		},
+		ObjectMeta: v1.ObjectMeta{
+			// imagejob will set node name as env when creating collector pod
+			Name: "imagecollector-" + os.Getenv("NODE_NAME"),
```
sozercan: do we need to generate this in case it already exists? imageCollector-NodeName-random

.gitignore
ashnamehrotra: The PROJECT file is from kubebuilder, since to add controllers/crds using kubebuilder the file needs to be present locally

pmengelbert: I think `PROJECT` should be checked in -- it wasn't there when I tried to create the collector crd and kubebuilder wouldn't work without the `PROJECT` file present

PROJECT
```diff
@@ -0,0 +1,21 @@
+domain: tutorial.kubebuilder.io
+layout:
+- go.kubebuilder.io/v3
+projectName: eraser
+repo: tutorial.kubebuilder.io/eraser
```
sozercan: shouldn't be tutorial i think?

sozercan: example: https://github.com/Azure/azure-workload-identity/blob/main/PROJECT

pkg/collector/collector.go
```diff
@@ -0,0 +1,167 @@
...
+type client struct {
+	images  pb.ImageServiceClient
+	runtime pb.RuntimeServiceClient
+}
+
+type Client interface {
+	listImages(context.Context) ([]*pb.Image, error)
+}
+
+func (c *client) listImages(ctx context.Context) (list []*pb.Image, err error) {
```
sozercan, line 41: looks like `listImages` is shared too?

Dockerfile
```diff
@@ -44,6 +45,20 @@ FROM --platform=$BUILDPLATFORM $ERASERBASEIMAGE as eraser
 COPY --from=eraser-build /workspace/out/eraser /
 ENTRYPOINT ["/eraser"]
 
+FROM builder AS collector-build
+
+ARG TARGETOS
```
sozercan: do we need to repeat these args? i think i added this to @pmengelbert's pr review

pkg/collector/collector.go
ashnamehrotra: I think we can't take the method away since the Client interfaces for eraser and collector are different. Would it be better to add a listImages to utils and call it from inside listImages from both collector/eraser?

cpuguy83:
I think this is missing the `ImageCollector` type definitions?

sozercan:
@cpuguy83 i think they got merged in https://github.com/Azure/eraser/pull/171

pkg/collector/collector.go
pmengelbert: You could factor out the body of the two `listImages` methods into a function that takes `c.images` (a `pb.ImageServiceClient`) as an argument. If that function was `util.ListImages`, then the bodies of these two methods would just be `return util.ListImages(c.images, ctx)`. Easier to maintain if you expect the two `listImages` methods to permanently remain identical.

pkg/utils/utils.go
```diff
@@ -0,0 +1,86 @@
+package pkg
```
pmengelbert: That's an odd name for a package. Usually it matches the directory it's in.  What about `package utils` instead?

pkg/eraser/eraser_test.go
```diff
@@ -64,7 +65,7 @@ func TestParseEndpointWithFallBackProtocol(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		p, a, e := parseEndpointWithFallbackProtocol(tc.endpoint, tc.fallbackProtocol)
+		p, a, e := util.ParseEndpointWithFallbackProtocol(tc.endpoint, tc.fallbackProtocol)
```
pmengelbert: Typically, tests will be in the same package as the package they're testing. The tests that are testing functions in `util` should be moved to `pkg/utils/utils_test.go` and have `package utils` (see my other comment about the package name).

pmengelbert (changes requested):

pmengelbert (approved):
LGTM

Dockerfile
```diff
@@ -2,8 +2,12 @@
 ARG BUILDERIMAGE="golang:1.17"
 
 ARG ERASERBASEIMAGE="gcr.io/distroless/static:latest"
+ARG COLLECTORBASEIMAGE="gcr.io/distroless/static:latest"
```
sozercan: looks like we are going to be adding same image in scanner too. can we just call these `STATICBASEIMAGE` and `STATICNONROOTBASEIMAGE` (feel free to change names)

pkg/utils/utils.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return "", "", fmt.Errorf("error while parsing: %w", err)
+	}
+
+	switch u.Scheme {
+	case "tcp":
+		return "tcp", u.Host, nil
+	case "unix":
+		return "unix", u.Path, nil
+
```
sozercan: nit: let's make this consistent, can we remove the new lines in between?

sozercan (approved):
a few suggestions, otherwise LGTM

PR #182, opened 2022-05-17
generate helm chart

### LINKED ISSUE #77: generate helm chart
- generate chart (example: https://github.com/open-policy-agent/gatekeeper/tree/master/cmd/build/helmify)
- add helm lint

it should look something like: https://github.com/Azure/azure-workload-identity/tree/main/third_party/open-policy-agent/gatekeeper/helmify

anmalkov:
Helm chart to avoid installation via static YAML file

fixes #77


sozercan:
@anmalkov  thanks for working on this! we want chart to be generated, not as static yaml. please see examples for Gatekeeper and Workload identity in #77 that use `helmify` to generate the chart.

sozercan:
Once we generate, it should be generated under `manifest_staging/charts`, which will then be copied in `charts` when we cut a release.

lee0c:
Hi @sozercan , I think we are having trouble understanding the benefits of `helmify` when there are no current files suited for use as the base for the YAML->Helm transformation. Do you then want the current `deploy/` files updated to have all the helm replacements templated out? Or are there files somewhere we are missing?

sozercan:
@lee0c `helmify` allows us to generate chart automatically, so we don't have to maintain a helm chart separately

this PR might help: https://github.com/Azure/azure-workload-identity/pull/66/files

anmalkov:
@sozercan but to generate automatically from what? I have seen that you do the replacements using the placeholders  in the files. Isn't it the main idea of the helm chart with values.yaml files? The crds can be easily copied using the step in the GitHub workflow. And all the rest will be done using the helm functionality. Am I missing something?

sozercan:
@anmalkov it's generated from kustomize, like https://github.com/Azure/eraser/tree/main/config/default

lee0c:
@sozercan Do you have a preference for where the helmify files live? We are considering `build/helmify`. (We know output files go in `manifest_staging`, this is just for the actual files helmify uses)

sozercan:
since we are using `helmify` from gatekeeper, let's add to `third_party/open-policy-agent/gatekeeper/helmify` and make sure to copy gatekeeper `LICENSE` and add a readme that details that we are re-using gatekeeper helmify (same as WI changes)

lee0c:
@sozercan So, we have been digging around on this. It doesn't look like either of the linked examples (`cmd/build/helmify` in Gatekeeper and `third_party/open-policy-agent/gatekeeper/helmify` in Workload Identity) actually use [helmify](https://github.com/arttor/helmify) in any way - they are just named helmify and run their own go code. Is this just a confusion due to naming collisions or are we supposed to be using a the linked helmify tool?

sozercan:
it's just a naming coincidence :) helmify we are talking about is here: https://github.com/open-policy-agent/gatekeeper/tree/master/cmd/build/helmify

joheec:
Currently blocked by this PR.

sozercan:
@joheec if you want to be unblocked asap, you can just take the static yamls from this and publish using https://github.com/stefanprodan/helm-gh-pages or https://github.com/helm/chart-releaser to your fork. once this pr is updated with auto-generation, we can use those

culster.yaml
```diff
@@ -0,0 +1,28 @@
+# this config file contains all config fields with comments
```
bdlb77, line 1: This `culster.yaml` file is committed in your PR

anmalkov:
Replaced by PR #191

PR #190, opened 2022-05-18
Windows docker build

### LINKED ISSUE #169: add windows support
support for windows nodes

example: 
https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/Makefile#L133-L139

vyta:
Builds eraser w/ windows target

Fixes #169 

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -53,14 +53,6 @@ jobs:
           CACHE_TO=type=gha,scope=eraser-node,mode=max \
           ERASER_IMG=${{ env.REGISTRY }}/${REPO}:${TAG}
 
-      - name: Build eraser for Windows
-        run: make docker-build-windows-all \
-          CACHE_FROM=type=gha,scope=eraser-node \
-          CACHE_TO=type=gha,scope=eraser-node,mode=max \
-          REGISTRY=${{ env.REGISTRY }} \
-          REPO=${REPO} \
-          VERSION=${TAG}
-
       - name: Log in to the GHCR
         uses: docker/login-action@v2
         with:
@@ -72,7 +64,6 @@ jobs:
         run: |
           make docker-push-manager MANAGER_IMG=${{ env.REGISTRY }}/${REPO}-manager:${TAG}
           make docker-push-eraser ERASER_IMG=${{ env.REGISTRY }}/${REPO}:${TAG}
-          make docker-push-windows-all REGISTRY=${{ env.REGISTRY }} REPO=${REPO} VERSION=${TAG}
 
       - name: Create GitHub release
         uses: "marvinpinto/action-automatic-releases@v1.2.1"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -57,8 +57,9 @@ jobs:
         run: make docker-build-windows-all \
           CACHE_FROM=type=gha,scope=eraser-node \
           CACHE_TO=type=gha,scope=eraser-node,mode=max \
-          OSVERSION=1809 \
-          ERASER_WIN_IMG=${{ env.REGISTRY }}/${REPO}-windows-${OSVERSION}-amd64:${TAG}
+          REGISTRY=${{ env.REGISTRY }} \
+          REPO=${REPO} \
+          VERSION=${TAG}
 
       - name: Log in to the GHCR
         uses: docker/login-action@v2
@@ -71,6 +72,7 @@ jobs:
         run: |
           make docker-push-manager MANAGER_IMG=${{ env.REGISTRY }}/${REPO}-manager:${TAG}
           make docker-push-eraser ERASER_IMG=${{ env.REGISTRY }}/${REPO}:${TAG}
+          make docker-push-windows-all REGISTRY=${{ env.REGISTRY }} REPO=${REPO} VERSION=${TAG}
 
       - name: Create GitHub release
         uses: "marvinpinto/action-automatic-releases@v1.2.1"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -53,6 +53,13 @@ jobs:
           CACHE_TO=type=gha,scope=eraser-node,mode=max \
           ERASER_IMG=${{ env.REGISTRY }}/${REPO}:${TAG}
 
+      - name: Build eraser for Windows
+        run: make docker-build-eraser-windows \
+          CACHE_FROM=type=gha,scope=eraser-node \
+          CACHE_TO=type=gha,scope=eraser-node,mode=max \
+          OSVERSION=1809 \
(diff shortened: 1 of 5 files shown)


vyta:
Looks like the e2e test job timed out. May we increase timeout duration and rerun?

// @helayoty  @sozercan

sozercan:
@pmengelbert  is this due to intermittent failure?

pmengelbert:
@sozercan yes, this is why I reopened the issue. It's been difficult to reproduce, so I'll add these logs to the issue tracking intermittent failures
@vyta

pmengelbert:
The test timed out twice in a row. It may actually be related to the changes, though I don't know why.

vyta:
I think I may know the issue, working on the fix. h/o thanks!

.github/workflows/release.yaml
```diff
@@ -53,6 +53,13 @@ jobs:
           CACHE_TO=type=gha,scope=eraser-node,mode=max \
           ERASER_IMG=${{ env.REGISTRY }}/${REPO}:${TAG}
 
+      - name: Build eraser for Windows
+        run: make docker-build-eraser-windows \
+          CACHE_FROM=type=gha,scope=eraser-node \
+          CACHE_TO=type=gha,scope=eraser-node,mode=max \
+          OSVERSION=1809 \
```
sozercan: I think we need to build multiple OSVERSIONs here

`ALL_OSVERSIONS.windows := 1809 1903 1909 2004 ltsc2022`

example: https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/Makefile#L154-L156

windows.Dockerfile
```diff
@@ -0,0 +1,37 @@
+# syntax=mcr.microsoft.com/oss/moby/dockerfile:1.3.1
+ARG BUILDERIMAGE="golang:1.17"
```
sozercan: we updated to 1.18 in another PR, let's match that  (golang:1.18-bullseye)

Makefile
```diff
@@ -114,6 +120,13 @@ build: generate fmt vet ## Build manager binary.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
+.PHONY: docker-buildx-builder
+docker-buildx-builder:
+	@if ! docker buildx ls | grep $(BUILDX_BUILDER_NAME); then \
+		docker buildx create --name $(BUILDX_BUILDER_NAME) --platform windows/amd64 --use; \
```
sozercan: i think we can skip `--platform windows/amd64` here but @cpuguy83 can correct me

Makefile
```diff
@@ -27,6 +31,8 @@ ifdef CACHE_FROM
 _CACHE_FROM := --cache-from $(CACHE_FROM)
 endif
 
+BUILDX_BUILDER_NAME ?= img-builder-win
```
sozercan: if we can skip --platform=windows/amd64, let's call this `img-builder`

Makefile
```diff
@@ -1,8 +1,12 @@
 VERSION := v0.1.0
 
+# Windows OS Version
+OSVERSION ?= 1809
```
sozercan, line 4: see my previous comment about multiple OS versions since it needs to match the host

.github/workflows/release.yaml
vyta: Cool, wasn't sure since it looked like there was only a single arch build for linux. Updated!

Makefile
vyta: Hmm, ok. Local testing kept throwing an "os unsupported" error w/o that. I updated anyway in case my error was specific to my local env.

Makefile
vyta: Kept this for local testing/one off windows builds. But there is an additional var w/ all oses

.github/workflows/release.yaml
sozercan: for Linux, we just added multi-arch https://github.com/Azure/eraser/pull/155 however, windows is a special case since it needs to match host version

Makefile
```diff
@@ -12,6 +18,8 @@ GOLANGCI_LINT_VERSION := 1.43.0
 
 PLATFORM ?= linux
 
+ALL_OSVERSIONS.windows := 1809 1903 1909 2004 ltsc2022
```
sozercan: @marosset do we still care for all these? can/should we drop any?

Makefile
```diff
@@ -132,6 +149,22 @@ docker-build-collector:
 docker-push-collector:
 	docker push ${COLLECTOR_IMG}
 
+docker-build-eraser-windows:
+	docker buildx build $(_CACHE_FROM) $(_CACHE_TO) --platform="windows/amd64" --output=type=docker --build-arg OSVERSION=${OSVERSION} -t ${REGISTRY}/${REPO}-windows-${OSVERSION}-amd64:${VERSION} -f windows.Dockerfile .
+
+docker-push-windows:
+	docker push ${REGISTRY}/${REPO}-windows-${OSVERSION}-amd64:${VERSION}
```
sozercan, line 160: @vyta is this going to work since we can't load windows images in linux?

vyta: Ah, you're right! Great catch. What's the best way to push the windows images? I suppose one option would be to update the output type of the buildx build to registry, but it appears registry login doesn't happen until after build step.

Makefile
marosset: we can drop 1903, 1909, 2004.
(20h2 is going out of support in Aug but I don't see it in the list)

Makefile
sozercan: that's right. this is the part that gets a little tricky, we need to push to a staging registry. for other projects, we use ADO to run this on our own agents/registry but we don't have that infra set up yet for eraser.

Makefile
marosset: (We probably just keep on not supporting 20H2)

Makefile
vyta: Gotcha. Perhaps we leave off the windows release in this PR and create a separate issue to track that work?

sozercan: SGTM. I am okay scoping it down to "support building windows images" and we can figure out the push/release at a later time once we have the infra set up for it

.github/workflows/test.yaml
```diff
@@ -104,6 +104,8 @@ jobs:
         run: make docker-build-manager CACHE_FROM=type=gha,scope=eraser-manager CACHE_TO=type=gha,scope=eraser-manager,mode=max
       - name: Build eraser
         run: make docker-build-eraser CACHE_FROM=type=gha,scope=eraser-node CACHE_TO=type=gha,scope=eraser-node,mode=max
+      - name: Build eraser for Windows
```
sozercan: let's remove this since we won't be able to test it yet and it will increase test duration

sozercan:
Closing this for now since we don't support Windows at this time due to no scanner available.

PR #192, opened 2022-05-18
Update go 18

### LINKED ISSUE #154: update ci to go 1.18
update to go 1.18

bdlb77:

This PR updates Golang to v1.18 throughout the Project. 

**Which issue(s) this PR fixes**:

Fixes #154 

Update Golang in:

- Dockerfiles
- go.mod 
- CI Pipeline

> Note: `go.mod` indirect requires are in own block form v1.17 and on.. `go.mod` was version v1.16 before this PR, so that is why there are many changes when running `go get -u && go mod tidy`


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/eraser
 
-go 1.16
+go 1.18
```
sozercan: we don't need to update this as we don't rely on go 1.18 features yet

sozercan:
i think we have a few Dockerfiles that needs to be updated, let's use `golang:1.18-bullseye` image for those

go.mod
bdlb77: Reverted this back to `1.16`

bdlb77:
> i think we have a few Dockerfiles that needs to be updated, let's use `golang:1.18-bullseye` image for those

Updated the images to `golang:1.18-bullseye` ... In both

`/Dockerfile` and `build/tooling/Dockerfile`

sozercan (approved):
Thank you!! LGTM

bdlb77:
Thank you @sozercan  for reviewing and feedback! I'm not able to Merge so I will wait for your team to do it :)

PR #204, opened 2022-05-19
Fix potential reconciliation bug

pmengelbert:
I was previously unaware that updating `imagejob.Status` would trigger a
reconcile.

This change is necessary because of the following situation:
- Currently, there are two calls to `r.updateJobStatus` in the
  `handleNewJob` method.
- Between those two calls, we are creating several pods, which will
  *also* trigger this reconciler, as will updates to those pods' status
  fields.
- It seems murky which will get processed by the reconciler first: the
  pod events or the second update to `imagejob.Status`.

This commit fixes that problem by only updating the `imagejob.Status`
once, and ensuring that this update pushes to the reconciliation queue
*before* any pods are created.


ashnamehrotra:
lgtm!

ashnamehrotra (approved):

Issue #203, opened 2022-05-19
Parallel image cleaning


eedorenko:
- Create Nginx and Redis deployment
- Delete Nginx and Redis deployment
- Start cleaning Nginx image
- Start cleaning Redis image before Nginx cleaning ends
- Make sure Nginx and Redis images are cleaned

Result: both images Nginx and Redis are cleaned successfully, but the first imagejob (nginx) and corresponding completed eraser pods are not deleted. 

The bug is reproducible by `deleteImagesParallel` test from #201 PR.

pmengelbert:
@eedorenko Thank you for pointing this out! PRs #204 and #205 are the first steps toward addressing this.

Probably the best option is to delete the ImageJob (which will also terminate the eraser pods) on ImageList update events, then create a new one, which will respond to the new list.

PR #205, opened 2022-05-20
Move imagejob deletion to imagelist controller

### LINKED ISSUE #200: Delete imagejobs in imagelist controller, not imagejob controller
- Delete imagejob in imagelist controller, not from the imagejob controller.
- Update Imagelist status field in the imagelist controller, not in the imagejob controller.

This is necessary because:
- The `imagejob` is being repurposed for management of `collector` pods, and not just `eraser` pods.
- The `imagecollector controller` needs to monitor the `imagejob` to determine when it is completed, in order to proceed to the scanning phase.
- With the current behavior of the `imagejob controller`, the `imagejob` will be cleaned up once the `eraser` pods are finished, preventing the `imagecollector controller` from properly monitoring the `imagejob`.
- This issue will block further progress on the `imagecollector controller`, so it should be prioritized.

**Environment:**
Not relevant

pmengelbert:
**This PR is based on #204, please see the relevant diff here:**
https://github.com/pmengelbert/eraser/compare/fix_potential_reconciliation_bug...pmengelbert:eraser:move_imagejob_deletion_to_imagelist_controller

- Delete imagejob in imagelist controller, not from the imagejob controller.
- Update Imagelist status field in the imagelist controller, not in the imagejob controller.

This is necessary because:
- The `imagejob` is being repurposed for management of `collector` pods, and not just `eraser` pods.
- The `imagecollector controller` needs to monitor the `imagejob` to determine when it is completed, in order to proceed to the scanning phase.
- With the current behavior of the `imagejob controller`, the `imagejob` will be cleaned up once the `eraser` pods are finished, preventing the `imagecollector controller` from properly monitoring the `imagejob`.
- This issue will block further progress on the `imagecollector controller`, so it should be prioritized.

Fixes #200 
Based on top of PR #204, that one should be merged first

Please enjoy, and see commit messages for rationale and details

### DIFF
--- a/controllers/imagelist/imagelist_controller.go
+++ b/controllers/imagelist/imagelist_controller.go
@@ -22,7 +22,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -99,20 +98,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	jobList := eraserv1alpha1.ImageJobList{}
-	err = r.List(ctx, &jobList, &client.ListOptions{LabelSelector: labels.SelectorFromSet(
-		map[string]string{
-			consts.ImageJobOwnerLabelKey: imageList.Name,
-		},
-	)})
+	err = r.List(ctx, &jobList)
 	if client.IgnoreNotFound(err) != nil {
 		return ctrl.Result{}, err
 	}
 
-	switch len(jobList.Items) {
+	items := util.FilterJobListByOwner(jobList.Items, *metav1.NewControllerRef(&imageList, imageList.GroupVersionKind()))
+
+	switch len(items) {
 	case 0:
 		return r.handleImageListEvent(ctx, &req, &imageList)
 	case 1:
-		job := jobList.Items[0]
+		job := items[0]
 
 		// If we got here because of a completed ImageJob:
 		if util.IsCompletedOrFailed(job.Status.Phase) {
--- a/controllers/imagelist/imagelist_controller.go
+++ b/controllers/imagelist/imagelist_controller.go
@@ -16,10 +16,12 @@ package imagelist
 import (
 	"context"
 	"flag"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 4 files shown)


controllers/imagelist/imagelist_controller.go
```diff
@@ -86,19 +92,55 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+}
+
+func (r *Reconciler) handleImageListEvent(ctx context.Context, req *ctrl.Request, imageList *eraserv1alpha1.ImageList) (ctrl.Result, error) {
 	job := &eraserv1alpha1.ImageJob{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: "imagejob-",
 			OwnerReferences: []metav1.OwnerReference{
 				*metav1.NewControllerRef(imageList, imageList.GroupVersionKind()),
 			},
+			Labels: map[string]string{consts.ImageJobOwnerLabelKey: imageList.Name},
```
cpuguy83: I think OwnerReference does what we need here?

controllers/imagelist/imagelist_controller.go
```diff
@@ -86,19 +92,54 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	)})
+	if client.IgnoreNotFound(err) != nil {
+		return ctrl.Result{}, err
+	}
+
+	switch len(jobList.Items) {
+	case 0:
+		return r.handleImageListEvent(ctx, &req, &imageList)
+	case 1:
+		return r.handleJobListEvent(ctx, &imageList, &jobList.Items[0])
```
pmengelbert: This needs additional logic. In addition to an ImageJob being moved to the `Completed` state, we will **also** reach this point when both of the following conditions are true:
1. Reconcile was triggered by an `UpdateEvent` to the `ImageList`
2. There is already an `ImageJob` in progress, and it is not yet complete.

This is related to #203 , and the best thing to do in that case is probably to requeue the `Imagelist`'s `UpdateEvent` for some reasonable time in the future, and to keep punting it until there are no `ImageJob`s in the running state.

cpuguy83 (changes requested):

controllers/imagelist/imagelist_controller.go
pmengelbert: Is there a way to select for that in the call to `r.List(...)` when we are listing the existing `ImageJob`s? I couldn't figure that out so I went with the labels instead. I am looking at the `sigs.k8s.io/controller-runtime/pkg/client.ListOptions` struct, and I don't see any option to select by `OwnerReference`.

The other option is to `List` all `ImageJob`s and filter them manually in the controller, is that the better way to go?

pmengelbert: Filtering client-side using `UID`s looks like the way to go, and more reliable than just using the owner's `Name`. Will mark resolved once tests pass

ashnamehrotra:
Can we also move the functionality from lines 287-310 into the imagelist controller? When we use the imagejob to schedule collector pods we won't need a reference to any imagelist

pmengelbert:
@ashnamehrotra Yes, that's a very good idea. I was running into problems with this yesterday and the conditional logic was getting messy. I'll work on that now

pmengelbert:
@ashnamehrotra done

sozercan (approved):
LGTM

cpuguy83 (approved):
LGTM

PR #207, opened 2022-05-23
Use correct base image for output artifacts

pmengelbert:
In `Dockerfile`, the terminal stages that produce our images are using
`BUILDPLATFORM` to determine the architecture of the base image before
pulling in the static binary. It should be `TARGETPLATFORM`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,10 +37,6 @@ RUN \
     --mount=type=cache,target=/go/pkg/mod \
     GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -ldflags '-w -extldflags "-static"' -o out/eraser ./pkg/eraser
 
-FROM --platform=$BUILDPLATFORM $STATICBASEIMAGE as eraser
-COPY --from=eraser-build /workspace/out/eraser /
-ENTRYPOINT ["/eraser"]
-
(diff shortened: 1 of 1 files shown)


cpuguy83 (approved):
LGTM

cpuguy83:
Deploy fails with:

```
 starting container process caused: exec: "/manager": stat /manager: no such file or directory
```

pmengelbert:
It's blowing up the CI, so there must be something wrong here. I'll look into it

Dockerfile
```diff
@@ -37,25 +37,24 @@ RUN \
...
 ENTRYPOINT ["/collector"]
 
+FROM --platform=$TARGETPLATFORM $STATICBASEIMAGE as eraser
+COPY --from=eraser-build /workspace/out/eraser /
+ENTRYPOINT ["/eraser"]
+
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM --platform=$BUILDPLATFORM $STATICNONROOTBASEIMAGE AS manager
-WORKDIR /
```
cpuguy83, line 58: Probably need this `WORKDIR`

cpuguy83 (approved):

Issue #208, opened 2022-05-27
Arguments to Eraser


ashnamehrotra:
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
- [x] Pass in repeat period
- [x] scanner image as arguments to Eraser for ImageCollector Controller.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Eraser version:
- Kubernetes version: (use `kubectl version`):

pmengelbert:
This will be resolved by #234

PR #213, opened 2022-05-31
ImageCollector Controller - Collector part (Fixed stacked changes)

### LINKED ISSUE #170: ImageCollector Controller - Collector part
ImageCollector Controller is responsible for scheduling ImageJob with collector pods, reading each individual imageCollector CR and creating a deduplicated list of images for ImageCollectorCR-shared.

**Environment:**

- Eraser version:

ashnamehrotra:
- initiates first reconcile for imagecollector controller
- periodically schedules an imagejob with collector pods
- once the job is finished, controller updates the shared imagecollector cr with deduplicated list of images in the spec & deletes the individual imagecollector crs

Fixes #170 


controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	)
+
+	if len(relevantJobs) > 1 {
+		return reconcile.Result{}, fmt.Errorf("More than one collector ImageJob scheduled")
+	}
+	if len(relevantJobs) == 0 {
+		if res, err := r.createImageJob(ctx, req, imageCollector); err != nil {
+			return res, err
+		}
+		return ctrl.Result{RequeueAfter: time.Second * 30}, nil
```
pmengelbert: I think the requeue time is too short, better something like every 10 minutes.  We should probably also have a flag to determine the timeout, which will help in changing this (since we requeue in many different places)

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	collectorImage = flag.String("collector-image", "ghcr.io/azure/collector:latest", "collector image")
```
pmengelbert, line 50: You will need to make changes to `config/manager/kustomization.template.yaml` to have this automatically supplied during deployment. You will also have to modify the `sed` command in the Makefile to do the substitution

controllers/imagejob/imagejob_controller.go
```diff
@@ -310,22 +310,23 @@ func (r *Reconciler) handleNewJob(ctx context.Context, imageJob *eraserv1alpha1.
...
-		log.Info("Started eraser pod on node", "nodeName", nodeName)
+		log.Info("Started "+containerName+" pod on node", "nodeName", nodeName)
 	}
 
 	return nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	log.Info("imagejob set up w manager")
```
pmengelbert: Spell out 'with`

controllers/imagejob/imagejob_controller.go
```diff
@@ -424,6 +425,7 @@ func copyAndFillTemplateSpec(templateSpecTemplate *corev1.PodSpec, env []corev1.
 	image.Env = append(env, image.Env...)
 	templateSpec.Volumes = append(volumes, templateSpec.Volumes...)
 	templateSpec.NodeName = nodeName
+	templateSpec.ServiceAccountName = "eraser-controller-manager"
```
pmengelbert: This should be supplied by the caller; remove here. It's already supplied from `imagecollector_controller`.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+				log.Info("could not create shared image collector")
+				return reconcile.Result{}, err
+			}
+		} else {
+			log.Info("could not get shared image collector")
+			return reconcile.Result{}, err
+		}
+	}
+
+	log.Info("xyz", "imageCollector", imageCollector)
```
pmengelbert: Helpful, but no longer necessary :)

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+		if _, err := r.handleJobDeletion(ctx, &relevantJobs[0]); err != nil {
+			return reconcile.Result{}, fmt.Errorf("Could not delete completed imagejob")
+		}
+	case eraserv1alpha1.PhaseFailed:
+		log.Info("failed phase")
+		if _, err := r.handleJobDeletion(ctx, &relevantJobs[0]); err != nil {
+			return reconcile.Result{}, fmt.Errorf("Could not delete failed imagejob")
+		}
+	default:
+		log.Info("should not reach this point for imagejob", relevantJobs[0])
```
pmengelbert: I think `log.Error` is called for here, since this should be unreachable.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+		return reconcile.Result{}, err
+	}
+
+	log.Info("Successfully created collector ImageJob", "job", job.Name)
+	return reconcile.Result{}, nil
+}
+
+func (r *Reconciler) updateSharedCRD(ctx context.Context, req ctrl.Request, imageCollector *eraserv1alpha1.ImageCollector) (ctrl.Result, error) {
+	imageCollectorList := &eraserv1alpha1.ImageCollectorList{}
+	if err := r.List(ctx, imageCollectorList); err != nil {
```
pmengelbert, line 295: This includes the shared ImageCollector resource, so we need to skip it in the loop

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	items := imageCollectorList.Items
+
+	var list []eraserv1alpha1.Image
+
+	for i := range items {
+		temp := items[i].Spec.Images
+		for _, img := range temp {
+			list = append(list, eraserv1alpha1.Image{Name: img.Name, Digest: img.Digest})
```
pmengelbert: No need to build a list here, just add to the map instead.

pmengelbert: Be sure to skip the shared ImageCollector:
```
if img.Name == "imagecollector-shared" {
    continue
}
```

"imagecollector-shared" should be actually defined as a constant, though.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	// else length is 1, so check job phase
+	switch phase := relevantJobs[0].Status.Phase; phase {
+	case eraserv1alpha1.PhaseCompleted:
+		log.Info("completed phase")
+		if res, err := r.updateSharedCRD(ctx, req, imageCollector); err != nil {
+			return res, err
+		}
+		if _, err := r.handleJobDeletion(ctx, &relevantJobs[0]); err != nil {
```
pmengelbert: This has problems. We're discarding the `ctrl.Result`, but the `ctrl.Result` has useful information in it (see line 219). There are two kinds of requeue we're looking at:
1. Requeue if there's a deletion delay
2. Requeue on a timer

we need to be able to handle both here.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/reconcile
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log.Info("ImageCollector Reconcile")
+
+	imageCollector := &eraserv1alpha1.ImageCollector{
+		TypeMeta:   metav1.TypeMeta{Kind: "ImageCollector", APIVersion: "eraser.sh/v1alpha1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "imagecollector-shared"},
```
pmengelbert: This should be a constant

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,324 @@
...
+
+		Spec: eraserv1alpha1.ImageCollectorSpec{Images: []eraserv1alpha1.Image{}},
+	}
+
+	if err := r.Get(ctx, types.NamespacedName{Name: "imagecollector-shared", Namespace: "default"}, imageCollector); err != nil {
```
pmengelbert: `"imagecollector-shared"` should be a constant

pmengelbert (changes requested):

controllers/imagecollector/imagecollector_controller.go
ashnamehrotra: The spec of the shared imagecollector is empty up until this point, do we still need to skip it? Assuming we are reset the spec to empty at the start of each collect/scan process?

controllers/imagecollector/imagecollector_controller.go
ashnamehrotra: I created an issue for the arguments to Eraser for the repeat period and scanner image. I can set this to a constant for now until we add that

controllers/imagecollector/imagecollector_controller.go
ashnamehrotra: I think there wouldn't be an imagecollector-shared image the nodes

controllers/imagecollector/imagecollector_controller.go
pmengelbert: I think it's worth doing because it clarifies intent, and is also a fail-safe. If, for whatever reason, the reconcile runs twice in the ten-minute repeat period, we don't want to modify the `imagecollector-shared` resource based on the current state of the `imagecollector-shared` resource. We just want to update it to reflect the most recent state.

controllers/imagecollector/imagecollector_controller.go
pmengelbert: ?

pmengelbert (changes requested):

controllers/imagecollector/imagecollector_controller.go
pmengelbert: I'm not sure I follow

ashnamehrotra: got confused since it said img.Name, but added a skip for the collector with items[i].Name

config/manager/manager.yaml
```diff
@@ -29,6 +29,7 @@ spec:
         - /manager
         args:
         - --leader-elect=false
+        - --job-cleanup-on-success-delay=10
```
pmengelbert: I think this is causing your CI failures on the e2e tests

pmengelbert: I think this is causing your CI failures on the e2e tests

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	imageCollectorList := &eraserv1alpha1.ImageCollectorList{}
+	if err := r.List(ctx, imageCollectorList); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	items := imageCollectorList.Items
+
+	// store images in map to remove duplicates
+	// map with key: sha id, value: name of image
+	idToTagListMap := make(map[string]string)
```
pmengelbert: nit: variable name is inaccurate. `idToNameMap` or `digestToNameMap`. Otherwise our future selves will be confused

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	imageCollector.Spec = eraserv1alpha1.ImageCollectorSpec{Images: combined}
+
+	if err := r.Update(ctx, imageCollector); err != nil {
+		log.Info("Could not update imageCollector spec")
+		return reconcile.Result{}, err
+	}
+
+	// delete individual image collector CRs
```
pmengelbert, line 335: I would put this in a separate method. This method is called `updateSharedCRD`, but it also deletes Kubernetes resources. That's pretty surprising behavior!

I'd split it into 2 functions, `updateSharedCR` and `deleteNodeCRs`. Then in the `Reconcile`, call them successively. That way the side-effects of each method are clear to the reader.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	// else length is 1, so check job phase
+	switch phase := relevantJobs[0].Status.Phase; phase {
+	case eraserv1alpha1.PhaseCompleted:
+		log.Info("completed phase")
+		if relevantJobs[0].Status.DeleteAfter == nil {
+			if res, err := r.updateSharedCRD(ctx, req, imageCollector); err != nil {
+				return res, err
+			}
+			relevantJobs[0].Status.DeleteAfter = util.After(time.Now(), *util.SuccessDelDelaySeconds)
+			if err := r.Status().Update(ctx, &relevantJobs[0]); err != nil {
```
pmengelbert, line 196: This status update is going to trigger another reconcile (it is an "update" event where `.Status.Phase` is either `Completed` or `Failed`). So you can just return with no requeue and no error.  Next time through the reconcile (after this update), when it gets back to this if statement, the other branch will be taken, because `.Status.DeleteAfter` will not be nil.

The reconcile resulting from this status update is going to happen no matter what (because of our predicate), so if we delete here we have a race condition. If the delete succeeds before the next reconcile's `List` operation, it will appear that there's no `ImageJob` at all and will create a new one (undesired behavior), and if the delete hasn't succeeded yet then it will end up back here and try to delete again, resulting in two calls to `Delete` when only one is necessary.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,342 @@
...
+func isNotFound(err error) bool {
+	if err != nil && client.IgnoreNotFound(err) == nil {
+		return true
+	}
+	return false
+}
+
+func (r *Reconciler) handleJobDeletion(ctx context.Context, job *eraserv1alpha1.ImageJob) (ctrl.Result, error) {
+	log.Info("start job deletion")
+	if job.Status.DeleteAfter == nil {
```
pmengelbert: I think we don't need to check this anymore, since we are checking before we call this method.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,342 @@
...
+			}
+		}
+		if res, err := r.handleJobDeletion(ctx, &relevantJobs[0]); err != nil || res.RequeueAfter > 0 {
+			return res, err
+		}
+	case eraserv1alpha1.PhaseFailed:
+		log.Info("failed phase")
+		if relevantJobs[0].Status.DeleteAfter == nil {
+			relevantJobs[0].Status.DeleteAfter = util.After(time.Now(), *util.ErrDelDelaySeconds)
+			if err := r.Status().Update(ctx, &relevantJobs[0]); err != nil {
```
pmengelbert, line 208: Ditto, see above.

config/manager/manager.yaml
ashnamehrotra: fixed the e2e, but looks like there's another issue testing locally with job cleanup delay and deploying an imagelist since it seems to hang on the deletion

controllers/imagelist/imagelist_controller.go
```diff
@@ -252,27 +272,17 @@ func (r *Reconciler) handleJobCompletion(ctx context.Context, imageList *eraserv
...
+		job.Status.DeleteAfter = util.After(time.Now(), *util.SuccessDelDelaySeconds)
+	} else if job.Status.Phase == eraserv1alpha1.PhaseFailed {
+		job.Status.DeleteAfter = util.After(time.Now(), *util.ErrDelDelaySeconds)
 	}
 
-	until := time.Until(job.Status.DeleteAfter.Time)
-	if until > 0 {
-		log.Info("Delaying imagejob delete", "job", job.Name, "deleteAter", job.Status.DeleteAfter)
-		return ctrl.Result{RequeueAfter: until}, nil
+	if err := r.Status().Update(ctx, job); err != nil {
```
pmengelbert: See my comment in `imagecollector_controller.go`, this is going to trigger another reconcile after we return, so we need to check the logic.

pmengelbert (changes requested):

pmengelbert:
@ashnamehrotra  Also needs to be resbased on main

config/manager/manager.yaml
pmengelbert: I think I found the source of this issue -- see the comments on deletion, we are queueing another reconcile when we update the DeleteAfter.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,342 @@
...
+			},
+		},
+		Spec: eraserv1alpha1.ImageJobSpec{
+			JobTemplate: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					RestartPolicy: corev1.RestartPolicyNever,
+					Containers: []corev1.Container{
+						{
+							Name:            "collector",
+							Image:           *collectorImage,
```
sozercan, line 263: where is this being set? we need something like https://github.com/Azure/eraser/blob/main/config/manager/kustomization.template.yaml#L18-L24

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	collectorImage = flag.String("collector-image", "ghcr.io/azure/collector:latest", "collector image")
+	log            = logf.Log.WithName("controller").WithValues("process", "imagecollector-controller")
+)
+
+const (
+	repeatPeriod    = time.Minute * 10
```
sozercan: can we make this configurable with a flag? let's make the default higher, perhaps 24 hours?

controllers/imagecollector/imagecollector_controller.go
sozercan: looks like @pmengelbert already mentioned this in https://github.com/Azure/eraser/pull/213/files#r885885202

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	relevantJobs := util.FilterJobListByOwner(
+		imageJobList.Items, metav1.NewControllerRef(imageCollector, schema.GroupVersionKind{
+			Group:   "eraser.sh",
+			Version: "v1alpha1",
+			Kind:    "ImageCollector",
+		}),
+	)
+
+	if len(relevantJobs) > 1 {
+		return reconcile.Result{}, fmt.Errorf("More than one collector ImageJob scheduled")
```
sozercan: nit:
```suggestion
		return reconcile.Result{}, fmt.Errorf("more than one collector ImageJobs are scheduled")
```

controllers/imagecollector/imagecollector_controller.go
pmengelbert:
```
diff --git a/Makefile b/Makefile
index 3092373..838dc4c 100644
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ lint: $(GOLANGCI_LINT) ## Runs go linting.
 ##@ Development

 manifests: __controller-gen ## Generates k8s yaml for eraser deployment.
-       @sed -e "s~ERASER_IMG~${ERASER_IMG}~g" config/manager/kustomization.template.yaml > config/manager/kustomization.yaml
+       @cat config/manager/kustomization.template.yaml | sed -e "s~ERASER_IMG~${ERASER_IMG}~g" | sed -e "s~COLLECTOR_IMG~${COLLECTOR_IMG}~g" > config/manager/kustomization.yaml
        docker run -v $(shell pwd)/config:/config -w /config/manager \
                k8s.gcr.io/kustomize/kustomize:v${KUSTOMIZE_VERSION} edit set image controller=${MANAGER_IMG}
        $(CONTROLLER_GEN) \
diff --git a/config/manager/kustomization.template.yaml b/config/manager/kustomization.template.yaml
index 859e4ea..73306af 100644
--- a/config/manager/kustomization.template.yaml
+++ b/config/manager/kustomization.template.yaml
@@ -20,5 +20,8 @@ patches:
     - op: add
       path: /spec/template/spec/containers/0/args/-
       value: --eraser-image=ERASER_IMG
+    - op: add
+      path: /spec/template/spec/containers/0/args/-
+      value: --collector-image=COLLECTOR_IMG
   target:
     kind: Deployment
```

config/manager/kustomization.template.yaml
```diff
@@ -22,3 +22,9 @@ patches:
       value: --eraser-image=ERASER_IMG
   target:
```
pmengelbert: Delete lines 23-25 here

pmengelbert (changes requested):

controllers/imagecollector/imagecollector_controller.go
pmengelbert: @sozercan We have an issue open #208 to add this. I'd rather not block merging this since it's the main blocker for several other things. WDYT?

ashnamehrotra: I added it as a flag with a 24h default, but we can leave that issue up to look more into it since the scanner image/type would need to be passed in later as well

sozercan: SGTM

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -0,0 +1,345 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	collectorImage = flag.String("collector-image", "ghcr.io/azure/collector:latest", "collector image")
+	log            = logf.Log.WithName("controller").WithValues("process", "imagecollector-controller")
+	repeatPeriod   = flag.Duration("repeat-period", time.Hour*24, "repeat period for collec/scan process")
```
sozercan:
```suggestion
	repeatPeriod   = flag.Duration("repeat-period", time.Hour*24, "repeat period for collect/scan process")
```

sozercan (approved):
LGTM

pmengelbert (approved):

PR #226, opened 2022-06-07
Separate vulnerable and failed images

### LINKED ISSUE #214: Modify ImageCollector resource to allow reporting of failures
During the scanning process, it is possible that the scanner is unable to scan a given image. In order to maintain visibility, I think `ImageCollectorStatus` should be modified to include a `Succeeded bool` field so that failures can be reported (rather than silently ignored).

**Environment:**
N/A

pmengelbert:
Resolves #214 , tweaks the design so that less logic is needed to handle vulnerable images, and less logic will be needed in the future when we deal with images that failed to scan.


ashnamehrotra (approved):
LGTM

PR #227, opened 2022-06-07
Modify collector

ashnamehrotra:
Modifies collector pods to report non-running images instead of all images. Shares code in util with eraser pods.
* Added another check in order to allow scanner to scan images by name rather than reporting only the digest.

Fixes #


pmengelbert (approved):
LGTM, thanks for the revisions.

PR #236, opened 2022-06-14
feat: Lock creation of new ImageJobs while ImageJob is in progress.

### LINKED ISSUE #180: Lock all collector operations (including scanning) while a collector operation is in progress.
The collect-scan-erase cycle will run on a timer, and we want to avoid starting a new cycle while an existing one is in progress. This could be controlled by mutex, or by some sort of locking mechanism for the kubernetes resources themselves.

pmengelbert:
Adds a locking mechanism to ImageJobs.

Fixes #180 

### DIFF
--- a/controllers/controller.go
+++ b/controllers/controller.go
@@ -16,12 +16,11 @@ var (
 	controllerLog      = ctrl.Log.WithName("controllerRuntimeLogger")
 	controllerAddFuncs []func(manager.Manager, util.KeyedLocker) error
 
-	keyMutex util.KeyedLocker
+	keyMutex = util.NewKeyedLocker()
 )
 
 func init() {
 	controllerAddFuncs = append(controllerAddFuncs, imagelist.Add, imagejob.Add, imagecollector.Add)
-	keyMutex = util.NewKeyedLocker()
 }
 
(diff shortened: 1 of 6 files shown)


pmengelbert:
This PR appears to have intermittent failures on the ImageList e2e tests, meaning those tests probably need to be refactored. The PR should also unlock the imagejob if it is deleted by some entity external to the ImageJob controller. (as is the case with the e2e tests)

pmengelbert:
Another problem here is that if we hold the lock until the ImageJob is `deleted`, the whole system is locked. This will be a problem if `--job-cleanup-on-success-delay` or `--job-cleanup-on-error-delay` are set to long values. This can be avoided if we unlock the ImageJob as soon as it's child jobs are completed.

However, if we unlock when the child jobs are completed instead of when the ImageJob is deleted, we run into a problem when the user deletes an ImageJob manually before the child jobs are completed: the lock won't be removed and the system ceases to function. This can only be fixed by restarting the manager.

pmengelbert:
Based on discussions, we will not implement this for the time being. Instead, we should note in the documentation that running multiple simultaneous imagejobs is not supported, and may result in unexpected behavior.

PR #261, opened 2022-06-22
feat: Exclusion list

### LINKED ISSUE #29: allow list for images and registries we don't want to remove
This can be done either as part of scanner, imagelist, or eraser

we can have a configmap that defines images to be in allow list (never remove even if they are vulnerable, etc)

- exclude specific images and tags (`'myregistry.azurecr.io/myImage:tag`)
- exclude all images in a registry (i.e., `myregistry.azurecr.io/*`)

ashnamehrotra:
Allows for "excluded" configmap to skip images and registries in eraser removal.

Fixes #29 

[Design Doc](https://docs.google.com/document/d/1ksUziJzNSVpwCagqmAzZOKllwbmZg0q2xJNUDUcQP2U/edit?usp=sharing)


pkg/eraser/eraser.go
```diff
@@ -198,6 +248,24 @@ func main() {
 		os.Exit(1)
 	}
 
+	// read excluded values from excluded configmap
+	data, err = os.ReadFile("/run/eraser.sh/excluded/excluded")
+	if err != nil {
+		if os.IsNotExist(err) {
+			log.Info("excluded configmap does not exist", "error: ", err)
+		} else {
+			log.Info("failed to read excluded values", "error: ", err)
```
pmengelbert: If the file is there but can't be read, I think `log.Error` is called for. We may even want to `os.Exit(1)` here, rather than proceed despite an IO error.

pkg/eraser/eraser.go
```diff
@@ -198,6 +248,24 @@ func main() {
...
+		}
+	}
+
+	// split images stored as comma separated values
+	vals := strings.Split(string(data), ",")
```
pmengelbert: As discussed, use a JSON blob in the configmap and unmarshal it here. Offers better format-checking, and the possibility of providing finer-grained controls in the future.

pkg/eraser/eraser.go
```diff
@@ -198,6 +248,24 @@ func main() {
 		os.Exit(1)
 	}
 
+	// read excluded values from excluded configmap
```
pmengelbert, line 284: I would put this in an `init()` function, since its purpose is to initialize a `var`.

pmengelbert (changes requested):

pkg/eraser/eraser.go
```diff
@@ -151,6 +162,45 @@ func removeImages(c Client, targetImages []string) error {
...
+	}
+
+	// check if img excluded by name
+	if len(idToTagListMap[img]) > 0 {
+		if _, contains := excluded[idToTagListMap[img][0]]; contains {
+			return true
+		}
+	}
+
+	r := regexp.MustCompile(`(\w)+\/\*\z`)
```
pmengelbert: `\w` won't match dashes or periods, which are both valid in image names. The image distribution spec's [full regex](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests) for image names is:
`[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*`.

So to make this work, it would be
```
r := regexp.MustCompile(`[a-z0-9]+([._-][a-z0-9]+)*\/\*\z`)
```

pmengelbert (changes requested):

test/e2e/imagelist/eraser_test.go
```diff
@@ -753,9 +753,123 @@ func TestRemoveImagesFromAllNodes(t *testing.T) {
...
+			}
+
+			return ctx
+		}).
+		Teardown(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			if err := util.DeleteEraserConfig(cfg.KubeconfigFile(), "eraser-system", "../test-data", "eraser_v1alpha1_imagelist.yaml"); err != nil {
+				t.Error("Failed to delete image list config ", err)
+			}
+			if err := util.DeleteImageListsAndJobs(cfg.KubeconfigFile()); err != nil {
+				t.Error("Failed to clean eraser obejcts ", err)
```
sozercan, line 863:
```suggestion
				t.Error("Failed to clean eraser objects", err)
```

pkg/eraser/eraser.go
```diff
@@ -151,6 +162,45 @@ func removeImages(c Client, targetImages []string) error {
...
+	if _, contains := excluded[img]; contains {
+		return true
+	}
+
+	// check if img excluded by name
+	if len(idToTagListMap[img]) > 0 {
+		if _, contains := excluded[idToTagListMap[img][0]]; contains {
+			return true
+		}
+	}
```
pmengelbert: An image can have multiple names associated with it. We need to consider all of them, not just the first, or we'll run afoul of the alias bug we fixed earlier.

```suggestion
	for _, imgName := range idToTagListMap[img] {
		if _, contains := excluded[imgName]; contains {
			return true
		}
	}
```

pkg/eraser/eraser.go
```diff
@@ -151,6 +162,45 @@ func removeImages(c Client, targetImages []string) error {
...
+			repo := strings.Split(key, "*")
+
+			// check if img is part of repo
+			if match2, _ := regexp.MatchString("^"+repo[0], img); match2 {
+				return true
+			}
+
+			// retrieve and check by name in the case img is digest
+			if len(idToTagListMap[img]) > 0 {
+				if match2, _ := regexp.MatchString(repo[0], idToTagListMap[img][0]); match2 {
```
pmengelbert: See above, we need to consider all names an image has

pkg/eraser/eraser.go
```diff
@@ -151,6 +162,45 @@ func removeImages(c Client, targetImages []string) error {
...
+
+	// look for excluded repository values
+	for key := range excluded {
+		// if excluded key ends in /*, check image with pattern match
+		if match := r.MatchString(key); match {
+			// store repository name
+			repo := strings.Split(key, "*")
+
+			// check if img is part of repo
+			if match2, _ := regexp.MatchString("^"+repo[0], img); match2 {
```
pmengelbert: This function swallows errors. It should return `(bool, error)` and return an error here if there's a problem.

Additionally, using a repo name as a regex could be problematic, because dots are valid in repo names, but they have a special meaning (a single wildcard character) in a regex. If you want to do this, you'll have to replace any `.` character with `\.`

pkg/eraser/eraser.go
pmengelbert: See above comments about swallowing errors and regex issues.

pkg/eraser/eraser.go
```diff
@@ -151,6 +162,45 @@ func removeImages(c Client, targetImages []string) error {
...
+	}
+
+	// check if img excluded by name
+	if len(idToTagListMap[img]) > 0 {
+		if _, contains := excluded[idToTagListMap[img][0]]; contains {
+			return true
+		}
+	}
+
+	r := regexp.MustCompile(`[a-z0-9]+([._-][a-z0-9]+)*\/\*\z`)
```
pmengelbert: One other point, in a regex you don't need to escape `/`. (I missed this in my first comment)

docs/QUICKSTART.md
```diff
@@ -161,6 +161,18 @@ Eraser can be used to [automatically](#automatically-cleaning-images) or [manual
     docker.io/library/alpine:3.7.3                                                                             application/vnd.docker.distribution.manifest.list.v2+json sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10 2.0 MiB   linux/386,linux/amd64,linux/arm/v6,linux/arm64/v8,linux/ppc64le,linux/s390x  io.cri-containerd.image=managed
     docker.io/library/alpine@sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10           application/vnd.docker.distribution.manifest.list.v2+json sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10 2.0 MiB   linux/386,linux/amd64,linux/arm/v6,linux/arm64/v8,linux/ppc64le,linux/s390x  io.cri-containerd.image=managed
     ```
+## Excluding images
```
sozercan:
```suggestion
## Excluding registries, repositories and images
```

docs/QUICKSTART.md
```diff
@@ -161,6 +161,18 @@ Eraser can be used to [automatically](#automatically-cleaning-images) or [manual
     docker.io/library/alpine:3.7.3                                                                             application/vnd.docker.distribution.manifest.list.v2+json sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10 2.0 MiB   linux/386,linux/amd64,linux/arm/v6,linux/arm64/v8,linux/ppc64le,linux/s390x  io.cri-containerd.image=managed
     docker.io/library/alpine@sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10           application/vnd.docker.distribution.manifest.list.v2+json sha256:8421d9a84432575381bfabd248f1eb56f3aa21d9d7cd2511583c68c9b7511d10 2.0 MiB   linux/386,linux/amd64,linux/arm/v6,linux/arm64/v8,linux/ppc64le,linux/s390x  io.cri-containerd.image=managed
     ```
+## Excluding images 
+Eraser can exclude registries (docker.io/library/*) and also specific images with a tag (azurecr.io/myImage:tag) or id (sha256:80f31da1ac7b312ba29d65080fd...) from its removal process.
```
sozercan:
```suggestion
Eraser can exclude registries (example, `docker.io/*`) and also specific images with a tag (example, `docker.io/library/ubuntu:18.04`) or digest (example, `sha256:80f31da1ac7b312ba29d65080fd...`) from its removal process.
```

docs/QUICKSTART.md
```diff
@@ -161,6 +161,18 @@ Eraser can be used to [automatically](#automatically-cleaning-images) or [manual
...
+
+To exclude any images or registries from the removal, create a configmap "excluded" in the eraser-system namespace with a JSON file holding the excluded images.
```
sozercan:
```suggestion
To exclude any images or registries from the removal, create a configmap named `excluded` in the eraser-system namespace with a JSON file holding the excluded images.
```

docs/QUICKSTART.md
sozercan: can we exclude all tags `docker.io/library/ubuntu:*`?

pkg/eraser/eraser.go
ashnamehrotra: good point!

docs/QUICKSTART.md
ashnamehrotra: it cant right now but I can add a check for that

pkg/eraser/eraser.go
```diff
@@ -153,6 +168,73 @@ func removeImages(c Client, targetImages []string) error {
...
+	// look for excluded repository values and names without tag
+	for key := range excluded {
+		// if excluded key ends in /*, check image with pattern match
+		if match := regexRepo.MatchString(key); match {
+			// store repository name
+			split := strings.Split(key, "*")
+			repo := strings.ReplaceAll(split[0], ".", `\.`)
+
+			// check if img is part of repo
+			if match, err := regexp.MatchString("^"+repo, img); match {
```
pmengelbert: I apologize for not coming to this conclusion sooner, especially since I've already asked you to change it; but since we are just checking whether or not `img` starts with `split[0]`, we can avoid regex entirely by using `strings.HasPrefix(img, split[0])` instead. In general terms, nothing but a string literal should be used as a regex, because it's error prone and hard for the reader to understand what is being compared.

Using `strings.HasPrefix` has the added bonus that we can probably remove the `error` return value.

pkg/eraser/eraser.go
```diff
@@ -153,6 +168,73 @@ func removeImages(c Client, targetImages []string) error {
...
+				return true, nil
+			} else if err != nil {
+				return false, err
+			}
+
+			// retrieve and check by name in the case img is digest
+			for _, imgName := range idToTagListMap[img] {
+				if match, err := regexp.MatchString(repo, imgName); match {
```
pmengelbert: ditto here

pkg/eraser/eraser.go
```diff
@@ -153,6 +168,73 @@ func removeImages(c Client, targetImages []string) error {
...
+			}
+		}
+
+		// if excluded key ends in :*, check image with pattern patch
+		if match := regexTag.MatchString(key); match {
+			// store image name
+			split := strings.Split(key, ":")
+			imagePath := strings.ReplaceAll(split[0], ".", `\.`)
+
+			if match, err := regexp.MatchString("^"+imagePath, img); match {
```
pmengelbert: and here

pkg/eraser/eraser.go
```diff
@@ -153,6 +168,73 @@ func removeImages(c Client, targetImages []string) error {
...
+				return true, nil
+			} else if err != nil {
+				return false, err
+			}
+
+			// retrieve and check by name in the case img is digest
+			for _, imgName := range idToTagListMap[img] {
+				if match, err := regexp.MatchString(imagePath, imgName); match {
```
pmengelbert: and here

pkg/eraser/eraser.go
```diff
@@ -209,6 +291,28 @@ func main() {
 		os.Exit(1)
 	}
 
+	// read excluded values from excluded configmap
+	data, err = os.ReadFile("/run/eraser.sh/excluded/excluded")
```
pmengelbert: use the string constant here

pkg/eraser/eraser.go
```diff
@@ -209,6 +291,28 @@ func main() {
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			log.Info("excluded configmap does not exist", "error: ", err)
+		} else {
+			log.Error(err, "failed to read excluded values")
+			os.Exit(1)
+		}
+	} else {
+		var result map[string][]string
```
pmengelbert: A struct is a little bit more type-safe, more extensible, and less error prone when we access the `Excluded` field. Then we can unmarshal into the struct.
```go
type ExclusionList struct {
	Excluded []string `json:"excluded"`
}
```
This struct should probably be in another package (`util` maybe) so it can be reused in the collector #274

pmengelbert (changes requested):

pmengelbert:
@ashnamehrotra we're very close, I think one more round of revisions then good to go.

pkg/eraser/eraser.go
ashnamehrotra: no worries, that makes sense!

pmengelbert (approved):
LGTM, just resolve the conflicts

PR #263, opened 2022-06-24
fix: Memory fix

### LINKED ISSUE #244: use volume mount for trivy for potential memory improvement
trivy scanner takes a lot of memory, perhaps using a volume mount as cache can help

pmengelbert:
Our usage of goroutines for concurrent scanning results in very large memory consumption, probably due to a goroutine leak. In order to fix this, I have removed the goroutines and instead run the scanner sequentially. The scan job takes longer, but the memory consumption is kept under acceptable limits.

Before this fix, on a fresh AKS cluster memory consumption would reach up to 3Gi. After this fix, it stays below 500Mi.

Fixes #244

### DIFF
--- a/pkg/scanners/trivy/trivy.go
+++ b/pkg/scanners/trivy/trivy.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"os"
 	"strings"
-	"sync"
 
 	machinerytypes "k8s.io/apimachinery/pkg/types"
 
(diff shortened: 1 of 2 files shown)


pkg/scanners/trivy/trivy.go
```diff
@@ -135,6 +134,13 @@ func main() {
 		os.Exit(generalErr)
 	}
 
+	if *enableProfile {
```
sozercan: can we move this to a separate pr?

sozercan (changes requested):

sozercan (approved):
lgtm

PR #264, opened 2022-06-24
feat: Enable pprof profiling

pmengelbert:
Useful for troubleshooting in the future

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,6 +18,9 @@ package main
 
 import (
 	"flag"
+	"fmt"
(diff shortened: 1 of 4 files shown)


sozercan (approved):

sozercan:
do we want to add this to other components (manager, eraser, collector, etc) too

pmengelbert:
> do we want to add this to other components (manager, eraser, collector, etc) too

Good idea

PR #294, opened 2022-07-06
feat: Enable helm args

pmengelbert:
Allows the user to supply arbitrary manager flags using the helm chart. Example:
```bash
$ helm install eraser manifest_staging/charts/eraser --set 'controllerManager.image.additionalArgs={--foo=bar,--non=sense}'
```


ashnamehrotra (approved):
lgtm thanks for explaining this!

PR #295, opened 2022-07-06
feat: Add controller flags to request/limit CPU/memory on scanner

### LINKED ISSUE #241: define cpu and memory requests and limits for scanner
Scanner does not have any requests and limits defined right now

pmengelbert:
Allows the user to set CPU or Memory requests/limits for the scanner pods.

Resolves #241

### DIFF
--- a/controllers/imagecollector/imagecollector_controller.go
+++ b/controllers/imagecollector/imagecollector_controller.go
@@ -380,10 +380,16 @@ func (r *Reconciler) createScanJob(ctx context.Context, collector *eraserv1alpha
 									"memory": resource.Quantity{
 										Format: resource.Format(*util.ScannerMemRequest),
 									},
+									"cpu": resource.Quantity{
+										Format: resource.Format(*util.ScannerCPURequest),
+									},
 								},
 								Limits: corev1.ResourceList{
 									"memory": resource.Quantity{
 										Format: resource.Format(*util.ScannerMemLimit),
+									},
+									"cpu": resource.Quantity{
+										Format: resource.Format(*util.ScannerCPULimit),
 									},
 								},
 							},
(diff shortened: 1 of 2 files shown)


controllers/util/util.go
```diff
@@ -13,6 +13,11 @@ import (
 var (
 	SuccessDelDelaySeconds = flag.Int64("job-cleanup-on-success-delay", 0, "Seconds to delay job deletion after successful runs. 0 means no delay")
 	ErrDelDelaySeconds     = flag.Int64("job-cleanup-on-error-delay", 86400, "Seconds to delay job deletion after errored runs. 0 means no delay")
+
+	ScannerCPULimit   = flag.String("scanner-cpu-limit", "", "limit on CPU usage for scanner pods spawned by the eraser manager")
+	ScannerCPURequest = flag.String("scanner-cpu-request", "", "minimum CPU request for scanner pods spawned by the eraser manager")
```
sozercan: can we define defaults for these too

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -389,6 +394,19 @@ func (r *Reconciler) createScanJob(ctx context.Context, collector *eraserv1alpha
 		},
 	}
 
+	resources := &scanJob.Spec.Template.Spec.Containers[0].Resources
```
sozercan: why are we handling these (cpu and mem) differently?

controllers/util/util.go
pmengelbert: Sure, what defaults do you recommend?

controllers/imagecollector/imagecollector_controller.go
pmengelbert: @sozercan because they have no default values associated (related to the above)

sozercan (approved):

PR #298, opened 2022-07-06
fix: dynamically generate list of tests

pmengelbert:
This PR dynamically generates the list of E2E tests. Before this change, whenever we added an e2e test we had to manually update the list of tests in both the Makefile and the test github workflow.

After this change, the list is determined dynamically (in bloth places) based on the directory structure. Each e2e test will be in its own directory in `test/e2e/tests`.  These get plugged into the CI's test matrix, and into the Makefile's list of e2e tests to run.


sozercan (approved):
Thank you! LGTM

PR #303, opened 2022-07-07
fix: Remove e2e test teardowns

### LINKED ISSUE #302: Remove Teardown sections from e2e tests
Each test has its own teardown section, which is no longer necessary. They used to be used in between tests, but because we are running the tests in parallel in the CI, we are now destroying the entire cluster after each test. We can safely remove the teardown sections from each of the e2e tests.

pmengelbert:
Each test has its own teardown section, which is no longer necessary.
They used to be used in between tests, but because we are running the
tests in parallel in the CI, we are now destroying the entire cluster
after each test. We can safely remove the teardown sections from each of
the e2e tests.

Resolves #302

### DIFF
--- a/test/e2e/tests/imagelist_change/eraser_test.go
+++ b/test/e2e/tests/imagelist_change/eraser_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 	"time"
 
-	eraserv1alpha1 "github.com/Azure/eraser/api/v1alpha1"
 	"github.com/Azure/eraser/test/e2e/util"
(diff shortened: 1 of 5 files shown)


sozercan:
@pmengelbert what happens if this is run locally? do devs need to teardown the clusters?

pmengelbert:
@sozercan These are just the "manual" teardowns that used to exist between tests when we ran them in sequence. But now the cluster gets completely torn down between every test and recreated by the next one. Devs won't need to tear down clusters.

sozercan (approved):

PR #307, opened 2022-07-08
chore: Update containerd to pass trivy scanner

pmengelbert:
Allows us to pass the trivy scanner in CI.


ashnamehrotra (approved):
lgtm

PR #309, opened 2022-07-08
docs: document eraser flags

### LINKED ISSUE #290: document eraser pod flags
Looks like we are missing these in the docs:

```
	successDelDelaySeconds = flag.Int64("job-cleanup-on-success-delay", 0, "Seconds to delay job deletion after successful runs. 0 means no delay")
	errDelDelaySeconds     = flag.Int64("job-cleanup-on-error-delay", 86400, "Seconds to delay job deletion after errored runs. 0 means no delay")
	successRatio           = flag.Float64("job-success-ratio", 1.0, "Ratio of successful/total runs to consider a job successful. 1.0 means all runs must succeed.")
```

ashnamehrotra:
Adds documentation for eraser pod flags with delays changed from seconds to duration.

Fixes #290 


docs/docs/eraser-flags.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Eraser Pod Flags
```
sozercan: `Customization` maybe?

docs/docs/eraser-flags.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Customization
+---
+
```
sozercan: let's add this is for eraser specifically. can these be configured through manager?

sozercan: if not, i think we need a way similar to how scanner args are passed through

ashnamehrotra: they can be passed in like the scanner args, but it applies to both eraser and collector right now. Should we remove the delay for collector pods?

pmengelbert (approved):

docs/docs/eraser-flags.md
ashnamehrotra: addressed eraser/collector pods in #313

docs/docs/eraser-flags.md
```diff
@@ -0,0 +1,11 @@
+---
```
sozercan: can we call the file customization.md too?

PR #318, opened 2022-07-12
fix: Add 'release' to options for semantic PR title

pmengelbert:


sozercan:
release action we are using doesn't have `release` as one of the options unfortunately https://github.com/marvinpinto/actions/blob/7132c40e0bf602ba0450c8c66337a421b94da3bb/packages/automatic-releases/src/utils.ts#L38

PR #308, opened 2022-07-08
feat: align time values

ashnamehrotra:
**What this PR does / why we need it**:
Change delay time arguments from seconds to duration to be consistent with repeat period.

**Which issue(s) this PR fixes** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #246

**Special notes for your reviewer**:

PR #325, opened 2022-07-14
fix: change seconds to duration

ashnamehrotra:
Update flag description to duration instead of seconds.

Fixes #


controllers/util/util.go
```diff
@@ -11,8 +11,8 @@ import (
 )
 
 var (
-	SuccessDel = flag.Duration("job-cleanup-on-success-delay", 0, "Seconds to delay job deletion after successful runs. 0 means no delay")
-	ErrDel     = flag.Duration("job-cleanup-on-error-delay", time.Hour*24, "Seconds to delay job deletion after errored runs. 0 means no delay")
+	SuccessDel = flag.Duration("job-cleanup-on-success-delay", 0, "Duration to delay job deletion after successful runs. 0 means no delay")
```
sozercan: What is the default unit?

ashnamehrotra: By default time.Duration is nanoseconds if the user doesn't provide a unit with their input. I can add that to the description as well.

sozercan: can we default to seconds?

sozercan: this can be a follow up if you want

ashnamehrotra: we changed it from seconds to duration in #308 for consistency. Should we revert it back?

cpuguy83 (approved):

PR #339, opened 2022-07-26
feat: scanner redesign

ashnamehrotra:
**What this PR does / why we need it**:
- Changes the design of the collect-scan-eraser pipeline. All containers run in one collector pod and used named pipes to communicate with each other. Scanner container is present only if scanner image provided. 
- Updates collector e2e tests that used imagecollector-shared
- Removes requirement of imagecollector-shared in reconcile 

**Which issue(s) this PR fixes** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #299,  #159

**Special notes for your reviewer**:
New design on last page of [design doc](https://docs.google.com/document/d/1Rz1bkZKZSLVMjC_w8WLASPDUjfU80tjV-XWUXZ8vq3I/edit#)

PR #348, opened 2022-08-04
fix: clean scanner logs output

### LINKED ISSUE #230: all trivy scanner logs should be json

Seeing non-json log lines, such as:
```
2022-06-08T01:58:32.588Z        INFO    Detected OS: debian
2022-06-08T01:58:32.588Z        INFO    Detecting Debian vulnerabilities...
2022-06-08T01:58:32.588Z        INFO    Number of language-specific files: 1
2022-06-08T01:58:32.588Z        INFO    Detecting gobinary vulnerabilities...
```

We should convert these to json if we can

ashnamehrotra:
Removes extra output logs produced by trivy library when scanning.

Fixes #230 

pr on top of #339


pkg/scanners/trivy/trivy.go
```diff
@@ -89,11 +90,17 @@ func main() {
...
 	ctx := context.Background()
 
+	var err error
+
 	if err := logger.Configure(); err != nil {
 		fmt.Fprintln(os.Stderr, "error setting up logger:", err)
 		os.Exit(generalErr)
 	}
 
+	if trivylogger.Logger, err = trivylogger.NewLogger(false, true); err != nil {
```
sozercan: looks like this disables trivy's logger, so we won't get those logs. instead of disabling it, can we replace `trivylogger.Log` with our own logger or create a new zap logger?

something like
```
logger, err := zap.NewProduction()
...
sugar := logger.Sugar()
trivylogger.Logger = sugar
```

sozercan (changes requested):
see comment

pkg/scanners/trivy/trivy.go
ashnamehrotra: (resolved)

pkg/scanners/trivy/trivy.go
```diff
@@ -89,11 +91,26 @@ func main() {
...
+	logger, err := zap.NewProduction()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "error setting up trivy logger:", err)
+		os.Exit(generalErr)
+	}
+
+	sugar := logger.Sugar()
+	trivylogger.Logger = sugar
+
+	if trivylogger.Logger, err = trivylogger.NewLogger(false, true); err != nil {
```
sozercan: i don't think we need this, isn't it going to overwrite the logger?

pkg/scanners/trivy/trivy.go
sozercan: yes, just the trivy logs and the scanner logs

pkg/scanners/trivy/trivy.go
```diff
@@ -89,11 +91,26 @@ func main() {
...
 
-	if err := logger.Configure(); err != nil {
+	var err error
+
+	if err = logger.Configure(); err != nil {
 		fmt.Fprintln(os.Stderr, "error setting up logger:", err)
 		os.Exit(generalErr)
 	}
 
+	logger, err := zap.NewProduction()
```
sozercan, line 102: let's add a comment that we are adding this because trivy logs are not json and we are doing this for json output

sozercan (approved):
LGTM

PR #366, opened 2022-08-11
feat: multiple exclusion

### LINKED ISSUE #338: exclustion configmap supports multiple config
To integrate Eraser on AKS, we want to keep some pre-defined system-level images in exclusion list and allow clients to define their ones in another configmap at the same time.

By searching the code base, looks like the exclusion configmap is now only loaded from `/run/eraser.sh/excluded/excluded`. 

Is this possible to support multiple exclusion configmap in future?

ashnamehrotra:
Allows for multiple exclusion configmaps, labeled with "eraser.sh/exclude.list=true".

Fixes #338

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -12,7 +12,6 @@ rules:
   verbs:
   - get
   - list
-  - watch
 - apiGroups:
   - ""
   resources:
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -8,6 +8,15 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - configmaps
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - nodes
   verbs:
   - get
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -10,7 +10,6 @@ rules:
   resources:
   - configmaps
   verbs:
-  - create
   - get
   - list
   - watch
--- a/controllers/imagecollector/imagecollector_controller.go
+++ b/controllers/imagecollector/imagecollector_controller.go
@@ -302,8 +302,8 @@ func (r *Reconciler) createImageJob(ctx context.Context, req ctrl.Request, argsC
 		return reconcile.Result{}, err
 	}
 
-	for i, container := range job.Spec.JobTemplate.Spec.Containers {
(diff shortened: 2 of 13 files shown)


config/rbac/role.yaml
```diff
@@ -5,6 +5,15 @@ metadata:
...
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - create
+  - get
+  - list
+  - watch
```
sozercan: do we need create or watch?

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -303,7 +290,25 @@ func (r *Reconciler) createImageJob(ctx context.Context, req ctrl.Request, argsC
 		job.Spec.JobTemplate.Spec.Containers = append(job.Spec.JobTemplate.Spec.Containers, scannerContainer)
 	}
 
-	err := r.Create(ctx, job)
+	configmapList := &corev1.ConfigMapList{}
+	if err := r.List(ctx, configmapList); err != nil {
```
sozercan: should we only look at eraser's ns?

controllers/imagejob/imagejob_controller.go
```diff
@@ -144,6 +144,7 @@ func checkNodeFitness(pod *corev1.Pod, node *corev1.Node) bool {
 
 //+kubebuilder:rbac:groups=eraser.sh,resources=imagejobs,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=eraser.sh,resources=imagejobs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=create;get;list;watch
```
sozercan: Should we add a namespace here? we don't want it hardcoded since charts can be deployed on any namespace though

config/rbac/role.yaml
ashnamehrotra: removed!

controllers/imagejob/imagejob_controller.go
ashnamehrotra: is there a way to pass a string into the kubebuilder comment? Im not sure if there is a way without hardcoding

sozercan: I was thinking it could be hardcoded here but we can override in the helm chart. is that possible?

ashnamehrotra: I tested it by hardcoding the namespace but I don't think it will be possible since eraser-manager-role doesn't apply to eraser-system which is the role we are using to access

sozercan (approved):
LGTM

PR #377, opened 2022-08-18
feat: upload logs

### LINKED ISSUE #280: upload logs on CI runs
upload manager/scanner/eraser logs so we can debug issues in the CI

ashnamehrotra:
Uploads manager and collector pod logs for e2e tests. Collector pod logs not included for collector-pipeline test since it deploys using manifest and imagelist-change test since it causes a timeout due to updating imagelist with delay. 

Fixes #280


cpuguy83:
Maybe you can write the logs to a file and have upload as a CI artifact? Should be a bit easier parse that way.
Maybe something like `TEST_LOGDIR` to determine where to write the files and then further namespace that on `<package name>/t.Name()/<pod name/manager>`?

Also seems like a lot of duplication here.
It seems like we could make a util function like:

```go
func CaptureLogs(t *testing.T, c Client, pods []*v1.Pod)
```
Maybe needs more arguments than that but just to give you an idea.

sozercan:
upload-artifact action that can help to push logs as a CI artifact: https://github.com/actions/upload-artifact

test/e2e/util/utils.go
```diff
@@ -391,6 +416,190 @@ func DeployEraserHelm(namespace string, args ...string) env.Func {
...
+	}
+
+	testName := strings.Split(t.Name(), "/")[0]
+
+	// get log output file path
+	path := filepath.Join(wd, TestLogDir, testName)
+
+	var file *os.File
+	if !fileExists(filepath.Join(path, manager.Name)) {
+		if err := os.MkdirAll(path, filemode); err != nil {
```
cpuguy83: This needs 0700 at least, probably 0755 would work best.
You can also just call `os.MkdirAll` without needing to check `fileExists`

test/e2e/util/utils.go
```diff
@@ -46,6 +50,8 @@ const (
 	FilterNodeSelector   = "kubernetes.io/hostname=eraser-e2e-test-worker"
 	FilterLabelKey       = "eraser.sh/cleanup.filter"
 	FilterLabelValue     = "true"
+	TestLogDir           = "../eraser_logs"
```
cpuguy83: This should be set in CI or something rather than a const.

So someone can do `make e2e LOGDIR=<path>`.
Can also set a default like `TEST_LOGDIR ?= $(PWD)/test_logs` or something in the Makefile.

cpuguy83 (changes requested):

test/e2e/util/utils.go
```diff
@@ -391,6 +416,169 @@ func DeployEraserHelm(namespace string, args ...string) env.Func {
...
+	var file *os.File
+	if err := os.MkdirAll(path, filemode); err != nil {
+		return err
+	}
+	_, err = os.Create(filepath.Join(path, manager.Name))
+	if err != nil {
+		return err
+	}
+
+	file, err = os.OpenFile(filepath.Join(path, manager.Name), os.O_APPEND|os.O_WRONLY, filemode)
```
cpuguy83: So before `filemode` was set to `0644` which is good for `os.OpenFile`.
It allows read/write for the user that created it, and read-only for users in the same group and everyone else.

Now `filemode is `0755` which gives execute permissions and other things that directories need.

I think I would just explicitly set `0644` in `OpenFile` and `0755` for `Mkdir` rather than using a constant.

cpuguy83 (changes requested):

cpuguy83:
Could you also add `.txt` as a file suffix on these logs? This way when trying to read them the OS knows how to open it.

cpuguy83 (approved):

PR #383, opened 2022-08-24
fix: add ReadHeaderTimeout for pprof profiling

ashnamehrotra:
Fixes new lint error:  _G114: Use of net/http serve function that has no support for setting timeouts (gosec)_

Fixes #

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	_ "net/http/pprof"
 	"os"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -70,7 +71,11 @@ func main() {
 
 	if *enableProfile {
 		go func() {
-			err := http.ListenAndServe(fmt.Sprintf("localhost:%d", *profilePort), nil)
+			server := &http.Server{
+				Addr:              fmt.Sprintf("localhost:%d", *profilePort),
+				ReadHeaderTimeout: 3 * time.Second,
```
jeremyrickard, line 76: I was curious about the addition of `ReadHeaderTimeout` here but I see the linter will also complain if that's omitted once you move to this method of creating the server. 

This LGTM, 3 seconds seems like a moderate timeout value...however this change essentially goes from [no timeouts](https://pkg.go.dev/net/http#Server)  to setting one for reading headers, so might be worth updating this PR title to include that so it shows up in the changelog/is easier to find if it becomes and issue later.

jeremyrickard (approved):
LGTM

PR #384, opened 2022-08-25
feat: separate code for scanner named pipe communication

### LINKED ISSUE #364: Scanner boilerplate code
Create boilerplate for scanner named pipes communication so that people can create and use their own scanners.

**Environment:**

- Eraser version:

ashnamehrotra:

Fixes #364 

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -295,8 +295,11 @@ func readConfigMap(path string) ([]string, error) {
 }
 
 func ReadCollectScanPipe(ctx context.Context) ([]eraserv1alpha1.Image, error) {
-	_, cancel := context.WithCancel(ctx)
-	defer cancel()
+	timer := time.NewTimer(time.Second)
+	if !timer.Stop() {
+		<-timer.C
+	}
+	defer timer.Stop()
 
 	var f *os.File
 	for {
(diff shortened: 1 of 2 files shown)


pkg/utils/utils.go
```diff
@@ -288,3 +293,55 @@ func readConfigMap(path string) ([]string, error) {
 
 	return images, nil
 }
+
+func ReadCollectScanPipe() ([]eraserv1alpha1.Image, error) {
```
cpuguy83: This function should support cancellation via context.

cpuguy83 (changes requested):

pkg/utils/utils.go
```diff
@@ -288,3 +293,58 @@ func readConfigMap(path string) ([]string, error) {
 
 	return images, nil
 }
+
+func ReadCollectScanPipe(ctx context.Context) ([]eraserv1alpha1.Image, error) {
+	_, cancel := context.WithCancel(ctx)
```
cpuguy83: For the cancellation here we shouldn't need to create a new context.
Instead the passed in context needs to be integrated into that for loop.

Something like:

```go
timer := time.NewTimer(time.Second)
if !timer.Stop() {
    <-timer.C
}
defer timer.Stop()
for {
    f, err = os.OpenFile(...)
   // error handling
  timer.Start(time.Second)
  select {
  case <-ctx.Done():
      return ctx.Err()
  case <-timer.C:
  }
}
```

This allows the caller to cancel the context in case it's taking too long to open the file.

cpuguy83 (changes requested):

pkg/utils/utils.go
ashnamehrotra: timer didn't have a Start() so I used Reset()  instead

cpuguy83 (approved):

PR #397, opened 2022-09-08
fix: multi-arch builds

sozercan:
Signed-off-by: Sertac Ozercan <sozercan@gmail.com>

This was preventing from targets setting up their own platform values

Fixes #


cpuguy83 (approved):
Seems like a bug that those ARGS that are outside a build stage are messing things up.
But we don't need them anyway (only useful when you need to reference the value in a FROM).

Confirmed this is now building aarch64 binaries when --platform=arm64.

PR #430, opened 2022-10-17
fix: show pods as running during collect-scan-erase pipeline

### LINKED ISSUE #395: Make collector pod's status clearer

When the collector pod is running, its status shows like "NotReady" which may confuse clients.

Can you make the pod's status as `Running` even though either scanner or eraser is not ready?

```
$ kubectl get pods --all-namespaces
NAMESPACE       NAME                                            READY   STATUS     RESTARTS      AGE
eraser-system   collector-aks-nodes-11597922-vmss000000-dpkqt   2/3     NotReady   0             19s
eraser-system   eraser-controller-manager-7ddc6c4fdb-z9lzp      1/1     Running    0             21s
```

pmengelbert:

The collector pods (containing the collector, scanner, and eraser containers) were showing as "NotReady" throughout the duration of the ImageJob. This was because when the collector pod finished its work, its process would end and therefore it would be seen as "Ready: false".

This PR resolves the issue by creating a new named pipe in both the collector and scanner containers. These pipes are opened immediately for reading, which blocks the cleanup of the collector and scanner containers. When the eraser pod finishes its work, it writes to and closes the collector and scanner completion named pipes. This stops the blocking on the read end, allowing all containers to shut down at roughly the same time. The end result is a transition from `Running` to `Completed` status for the collector pods.

Fixes #395

### DIFF
--- a/manifest_staging/deploy/eraser.yaml
+++ b/manifest_staging/deploy/eraser.yaml
@@ -4417,8 +4417,8 @@ spec:
       containers:
       - args:
         - --leader-elect=false
-        - --eraser-image=eraser:test
-        - --collector-image=collector:test
+        - --eraser-image=ghcr.io/azure/eraser:v0.4.0
+        - --collector-image=ghcr.io/azure/collector:v0.4.0
         - --scanner-image=ghcr.io/azure/eraser-trivy-scanner:v0.4.0
         command:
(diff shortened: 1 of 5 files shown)


sozercan (approved):
LGTM

PR #445, opened 2022-10-28
ci: Use locally built images during testing

### LINKED ISSUE #444: Use pullPolicy: Never in CI testing
Before PR #443 , the wrong image for the controller manager was being pulled into the kind cluster. The image built from the source was being ignored. This should be impossible, so we should set all images to pullPolicy: Never in the CI, which would cause the tests to fail when the wrong image is specified. This can be done using helm flags for most of the e2e tests. Deployments from the manifest will have to be considered separately.

pmengelbert:
See issue #444 

Fixes #444

### DIFF
--- a/controllers/util/util.go
+++ b/controllers/util/util.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"flag"
+	"fmt"
 	"time"
 
 	eraserv1alpha1 "github.com/Azure/eraser/api/v1alpha1"
@@ -25,9 +26,9 @@ var (
 	EraserImage = flag.String("eraser-image", "ghcr.io/azure/eraser:latest", "eraser image")
 	EraserArgs  = utils.MultiFlag([]string{})
 
-	EraserPullPolicy    corev1.PullPolicy
-	CollectorPullPolicy corev1.PullPolicy
-	ScannerPullPolicy   corev1.PullPolicy
+	eraserPullPolicy    = flag.String("eraser-pull-policy", string(corev1.PullIfNotPresent), "pull policy for eraser image")
+	collectorPullPolicy = flag.String("collector-pull-policy", string(corev1.PullIfNotPresent), "pull policy for collector image")
+	scannerPullPolicy   = flag.String("scanner-pull-policy", string(corev1.PullIfNotPresent), "pull policy for scanner image")
 )
 
 const (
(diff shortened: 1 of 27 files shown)


main.go
```diff
@@ -50,8 +52,11 @@ var (
 	enableLeaderElection = flag.Bool("leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	enableProfile = flag.Bool("enable-pprof", false, "enable pprof profiling")
-	profilePort   = flag.Int("pprof-port", 6060, "port for pprof profiling. defaulted to 6060 if unspecified")
+	enableProfile       = flag.Bool("enable-pprof", false, "enable pprof profiling")
+	profilePort         = flag.Int("pprof-port", 6060, "port for pprof profiling. defaulted to 6060 if unspecified")
+	eraserPullPolicy    = flag.String("eraser-pull-policy", string(corev1.PullIfNotPresent), "pull policy for eraser image")
```
sozercan: can we handle this in helm chart instead of being a flag? perhaps similar to `AdditionalArgs` or a configmap we can read?

pmengelbert:
It's complicated because the configuration for the eraser/scanner/collector containers are set in code, and not in the helm chart. For example to get args to the scanner from the helm chart we have to set --scanner-arg on the manager.
________________________________

main.go
pmengelbert: It's complicated because the configuration for the eraser/scanner/collector containers are set in code, and not in the helm chart. For example to get args to the scanner from the helm chart we have to set --scanner-arg on the manager.

pmengelbert: The tests use `--set controllerManager.image.additionalArgs={...}` to propagate these values to the right place. Maybe we should consider a system-wide configmap that the manager looks at when it boots up, so that we don't clutter up the arguments like this? @sozercan if you agree I'll open an issue for that.

sozercan: I am okay with configmap compared to cluttering arguments (which will get worse over time). Can we fix the e2e issue first then come up with a design for configmap solution?

pmengelbert: The configmap issue is being tracked by #446. I'm addressing the e2e issues

sozercan:
@pmengelbert do we want to rename the pr?

main.go
```diff
@@ -80,6 +85,21 @@ func main() {
 		}()
 	}
 
+	pullPolicies := map[string]*corev1.PullPolicy{
+		*eraserPullPolicy:    &util.EraserPullPolicy,
+		*collectorPullPolicy: &util.CollectorPullPolicy,
+		*scannerPullPolicy:   &util.ScannerPullPolicy,
+	}
```
pmengelbert: - [x] Flip these, the string keys could all have the same value.

test/e2e/util/utils.go
```diff
@@ -737,7 +767,7 @@ func CreateExclusionList(namespace string, list pkgUtil.ExclusionList) env.Func
 		excluded := corev1.ConfigMap{
 			ObjectMeta: metav1.ObjectMeta{
 				GenerateName: "excluded",
-				Namespace:    EraserNamespace,
+				Namespace:    TestNamespace,
```
pmengelbert: This function accepts a `namespace` parameter, so this should be replaced with `namespace`

test/e2e/util/utils.go
```diff
@@ -747,7 +777,7 @@ func CreateExclusionList(namespace string, list pkgUtil.ExclusionList) env.Func
 
 		cMap := corev1.ConfigMap{}
 		err = wait.For(func() (bool, error) {
-			err := c.Resources().Get(ctx, excluded.Name, EraserNamespace, &cMap)
+			err := c.Resources().Get(ctx, excluded.Name, TestNamespace, &cMap)
```
pmengelbert: here too

controllers/util/util.go
```diff
@@ -24,6 +25,10 @@ var (
 
 	EraserImage = flag.String("eraser-image", "ghcr.io/azure/eraser:latest", "eraser image")
 	EraserArgs  = utils.MultiFlag([]string{})
+
+	eraserPullPolicy    = flag.String("eraser-pull-policy", string(corev1.PullIfNotPresent), "pull policy for eraser image")
```
sozercan, line 29: i thought we wanted to move these parts out of this PR?

pmengelbert: @sozercan I misunderstood. Prior to this PR, the Pull Policy was hard coded so there is no option for the user (or tests) to change it. I thought you meant "leave the flags for now, then implement the configmap later and remove the flags". We need either the flags or the configmap for this PR. Should I put this one on hold until the configmap is implemented?

pmengelbert:
Closing this. See #458

PR #458, opened 2022-11-03
test: prevent pulling default images in e2e test

### LINKED ISSUE #444: Use pullPolicy: Never in CI testing

pmengelbert:

The file `test/e2e/test-data/helm-empty-values.yaml` has been introduced to prevent regression in the tests. A situation arose in which the helm keys were wrong in the e2e tests, resulting in helm plugging in the default values for image repository and tag. This resulted in false positives because the cluster used in the e2e test would pull in the default image from a registry.

For all e2e tests using helm, this file should be provided on the command line using `helm install -f`. This ensures that the default values are never used. The correct values will be supplied by `--set` flags further to the right on the command line.

In that file, randomized names are used to guarantee that if the wrong helm keys are used, the test will fail.

Fixes #444 

### DIFF
--- a/test/e2e/test-data/helm-empty-values.yaml
+++ b/test/e2e/test-data/helm-empty-values.yaml
@@ -1,0 +1,60 @@
+# This file exists to prevent regression in the tests. A situation arose in
+# which the helm keys were wrong in the e2e tests, resulting in helm receiving
+# the default values for image repository and tag. This resulted in false
+# positives because the cluster used in the e2e test would pull in the default
+# image from a registry.
+#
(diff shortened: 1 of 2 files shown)


sozercan (approved):
LGTM

PR #476, opened 2022-11-14
fix: Remove tooling containers created in Makefile

pmengelbert:
The Makefile has several `docker run ...` commands which do not remove the containers they create. This pollutes the user's system with hundreds of stopped containers.


sozercan (approved):
LGTM

PR #477, opened 2022-11-14
fix: image reference logic

### LINKED ISSUE #469: Fix image reference resolution logic
When querying the Container Runtime Interface for a list of images present on the node, it provides several different ways of identifying an image:
1. ImageID `string`
1. RepoTags (repo name plus tag, ex. `docker.io/library/alpine:latest`) `[]string`
1. RepoDigests (repo name plus digest, ex. `docker.io/library/alpine@sha256:dda987...`) `[]string`

Each of these has its own pros and cons which make them vary in usefulness during each of the many steps in the collect->scan->erase pipeline.

All of them uniquely identify an image (one RepoTag refers to one image, one RepoDigest refers to one image). But ImageID is the only one that is truly one-to-one in both directions. One Image will have only one ImageID, whereas it may have multiple RepoTags or RepoDigests.

## ImageId
**Pros:**
- Present in CRI ListImageResponse
- Present in CRI ListContainersResponse
- Useful at the time of filtering (all images on the node will have one, running containers will identify their base image with one)
- Useful at the time of erasure (handled by the CRI)

**Cons:**
- Impossible to search for ImageID using containerd (and therefore trivy), making them useless during scanning
- Easily confused with Digests
- Not strictly content-addressable, and may depend on the name of the image (i.e. docker.io/library/nginx:latest and abc.azurecr.io/nginx:latest may have different ImageIds, even when the image content is the same)
- Similar, not a strictly unique identifier of an image

## RepoTags
**Pros:**
- Avoids ID/Digest dichotomy
- Human readable, most commonly used by end-user
- Searchable by containerd/trivy

**Cons:**
- Aliasing can happen: two names refer to the same image, but telling the CRI to remove one name leaves the image present under the other name
- One Image may have many RepoTags, making them less useful at filter time (an image `A` may be running as a container under an alias `B`, thus the logic does not recognize that `A` is running)
- Not always present in CRI ListImagesResponse
- Not always present in CRI ListContainersResponse

## RepoDigests
**Pros:**
- Uniquely identifies an image when the Repo part is stripped (leaves only the Digest)
- Digest is more stable than ImageID: strictly content-addressable, does not change when name changes
- Recognized by containerd/trivy

**Cons:**
- One Image may have multiple digests (I am not fully sure how, but I think for different architectures?)
- Not always present in CRI ListImagesResponse
- Not always present in CRI ListContainersResponse

## Moving forward

We currently discard too much of this information, and we do so almost immediately after getting the list of images from the CRI:
- We currently only preserve one RepoTag when there may be several
- We are confusing ImageID with Digest, therefore making it impossible for containerd/trivy to find images without RepoTags.
- Related, we essentially discard Digests, preserving only the ImageID (but calling it a Digest)

**We need to preserve all of this information to make the resolution at filter time, scan time, and eraser time to work properly.**

**Environment:**

- Eraser version: `v0.5.0`

pmengelbert:
Resolves #469 by preserving imageID, digest, and name information from images throughout the collect->scan->erase pipeline.

### DIFF
--- a/api/v1alpha1/imagejob_types.go
+++ b/api/v1alpha1/imagejob_types.go
@@ -22,8 +22,9 @@ import (
 )
 
 type Image struct {
-	Digest string `json:"digest"`
-	Name   string `json:"name,omitempty"`
+	ImageID string   `json:"image_id"`
+	Names   []string `json:"names,omitempty"`
+	Digests []string `json:"digests,omitempty"`
 }
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
--- a/pkg/collector/helpers.go
+++ b/pkg/collector/helpers.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"strings"
 
 	eraserv1alpha1 "github.com/Azure/eraser/api/v1alpha1"
 	util "github.com/Azure/eraser/pkg/utils"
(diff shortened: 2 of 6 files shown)


sozercan (approved):
LGTM

PR #493, opened 2022-12-01
feat: pull secret

### LINKED ISSUE #492: Add global imagePullSecrets setting
Add support for a global imagePullSecrets setting to allow all images to pull from private registries using secret.

**Environment:**

- Eraser version: v0.5.0
- Kubernetes version: (use `kubectl version`): 1.24.6

pmengelbert:
Allows the user to set a set of global `imagePullSecrets`, for retrieving eraser component images (eraser, manager, scanner, collector) from private registries.

To use:
- Create a docker-registry secret (ex `mysecret`) in the same namespace in which eraser will be deployed
- Deploy with helm, by using the flag `--set-json eraserPullSecrets='[ { "name": "mysecret" } ]'`

Fixes #492 

Have fun and happy reviewing

### DIFF
--- a/manifest_staging/charts/eraser/values.yaml
+++ b/manifest_staging/charts/eraser/values.yaml
@@ -43,3 +43,4 @@ scanner:
     args: []
 
 nameOverride: ""
+eraserPullSecretName: ""
--- a/manifest_staging/charts/eraser/values.yaml
+++ b/manifest_staging/charts/eraser/values.yaml
@@ -43,4 +43,4 @@ scanner:
     args: []
 
 nameOverride: ""
-eraserPullSecretName: ""
+eraserPullSecrets: []
--- a/manifest_staging/charts/eraser/values.yaml
+++ b/manifest_staging/charts/eraser/values.yaml
@@ -43,4 +43,4 @@ scanner:
     args: []
 
 nameOverride: ""
-eraserPullSecrets: []
+imagePullSecrets: []
--- a/test/e2e/tests/helm_pull_secret/eraser_test.go
+++ b/test/e2e/tests/helm_pull_secret/eraser_test.go
@@ -1,0 +1,90 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/Azure/eraser/test/e2e/util"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/e2e-framework/klient/wait"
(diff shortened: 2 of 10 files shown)


manifest_staging/charts/eraser/values.yaml
```diff
@@ -43,3 +43,4 @@ scanner:
     args: []
 
 nameOverride: ""
+eraserPullSecretName: ""
```
sozercan: should we do this per image instead of global so it'll be more future proof

sozercan: do we want it a list similar to https://github.com/Azure/azure-workload-identity/blob/main/charts/workload-identity-webhook/values.yaml#L13

pmengelbert: It's a pod-level setting (can't be done per image if the containers share a pod), so collector/scanner/eraser will always have the same set of pullSecrets. If necessary we could let the user set a different secret for the manager.

pmengelbert: We can do a list instead, yes. It makes the logic a little more complicated because those values have to make their way to the code that generates the collector, etc pods.

test/e2e/tests/helm_pull_secret/eraser_test.go
```diff
@@ -0,0 +1,90 @@
...
+				t.Errorf("could not list pods: %v", err)
+			}
+
+			var ls2 corev1.PodList
+			err = c.Resources().List(ctx, &ls2, func(o *metav1.ListOptions) {
+				o.LabelSelector = labels.SelectorFromSet(map[string]string{"control-plane": "controller-manager"}).String()
+			})
+
+			items := append(ls.Items, ls2.Items...)
+			if len(items) != 4 {
```
sozercan: nit: let's make this a const so it's easier to track and update the string too

```suggestion
			if len(items) != expectedPods {
```

manifest_staging/charts/eraser/values.yaml
pmengelbert: Just pushed changes which make this a list of objects `{ "name": "mysecret" }`

test/e2e/tests/helm_pull_secret/eraser_test.go
pmengelbert: pushed changes fixing this

manifest_staging/charts/eraser/values.yaml
```diff
@@ -43,3 +43,4 @@ scanner:
     args: []
 
 nameOverride: ""
+eraserPullSecrets: []
```
sozercan: nit: any reason we want to use eraser?
```suggestion
imagePullSecrets: []
```

pmengelbert: done

sozercan (approved):
Thank you! LGTM

PR #499, opened 2022-12-05
chore: update CRI client runtime version to v1 (test with multiple k8s versions)

pmengelbert:


sozercan:
@cpuguy83 suggested, "If we make it so we can fallback to v1alpha2 in case v1 is not available that would work."

api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -28,6 +28,16 @@ import (
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
 func (in *Image) DeepCopyInto(out *Image) {
 	*out = *in
+	if in.Names != nil {
```
sozercan, line 31: why did this get updated?

pmengelbert: `make generate` hasn't been run in a while, and it got run as part of `make test`. The changes in #477 caused this, as it changed the definition of the `Image` type. We should probably implement a codegen check for the CI, similar to the manifest check we already have

pmengelbert: Just added that check to the CI for this PR

sozercan: that's what i was thinking. awesome, thank you!

pkg/utils/utils.go
```diff
@@ -40,9 +40,11 @@ const (
 	EnvEraserContainerRuntime = "ERASER_CONTAINER_RUNTIME"
 )
 
-type ExclusionList struct {
-	Excluded []string `json:"excluded"`
-}
+type (
```
cpuguy83: Why this change? This is pretty abnormal for go.

pmengelbert: Slipped through (there was other stuff in there before and I moved it out).

cpuguy83 (approved):

Issue #503, opened 2022-12-05
collector pod is not scheduling for all nodes


amrashwan:
**What steps did you take and what happened:**
collector pod is not scheduling for all nodes 
Please create a property/settings in the chart to pass the toleration to the collector pod

Example of the errors
52s         Normal    Started                pod/collector-aks-system-18979628-vmss000000-fsghv   Started container trivy-scanner
56s         Warning   TaintToleration        pod/collector-aks-xinfra-11501615-vmss000000-x4r89   Predicate TaintToleration failed
56s         Normal    TaintManagerEviction   pod/collector-aks-xinfra-11501615-vmss000000-x4r89   Marking for deletion Pod xinfra/collector-aks-xinfra-11501615-vmss000000-x4r89
56s         Normal    TaintManagerEviction   pod/collector-aks-xinfra-11501615-vmss000001-htknt   Marking for deletion Pod xinfra/collector-aks-xinfra-11501615-vmss000001-htknt
56s         Warning   TaintToleration        pod/collector-aks-xinfra-11501615-vmss000001-htknt   Predicate TaintToleration failed
56s         Normal    TaintManagerEviction   pod/collector-aks-xinfra-11501615-vmss000002-l2mct   Marking for deletion Pod xinfra/collector-aks-xinfra-11501615-vmss000002-l2mct
56s         Warning   TaintToleration        pod/collector-aks-xinfra-11501615-vmss000002-l2mct   Predicate TaintToleration failed
56s         Normal    TaintManagerEviction   pod/collector-aks-xinfra-11501615-vmss000003-bxsrb   Marking for deletion Pod xinfra/collector-aks-xinfra-11501615-vmss000003-bxsrb
56s         Warning   TaintToleration        pod/collector-aks-xinfra-11501615-vmss000003-bxsrb   Predicate TaintToleration failed
56s         Normal    TaintManagerEviction   pod/collector-aks-xinfra-11501615-vmss000004-tlgwz   Marking for deletion Pod xinfra/collector-aks-xinfra-11501615-vmss000004-tlgwz
56s         Warning   TaintToleration        pod/collector-aks-xinfra-11501615-vmss000004-tlgwz   Predicate TaintToleration failed

**What did you expect to happen:**
To schedule collector pods to all linux nodes on the cluster

**Environment:**

- Eraser version: 0.5.0
- Kubernetes version: (use `kubectl version`): 1.24.6

pmengelbert:
[...]

PR #508, opened 2022-12-06
fix: add default tolerations for collector/eraser pods

pmengelbert:
Issue #503 points out that pods will not be scheduled on nodes with taints, because we do not set any tolerations on the collector or eraser pods.

By contrast, daemonsets have default tolerations that enable them to be scheduled on every node regardless of taints. This PR applies those defaults to the collector and eraser pods.

Enjoy life, it's nice.


sozercan:
@pmengelbert we will still be able to exclude nodes by label though, is that correct?

sozercan (approved):
LGTM

pmengelbert:
> @pmengelbert we will still be able to exclude nodes by label though, is that correct?

Yes, and the e2e tests verify that this doesn't break it. Merging now

PR #520, opened 2022-12-08
test: create additional test using containerd v1.6.x

pmengelbert:
By default the e2e test framework will install kind v1.12.0 (if not already installed), which uses containerd version 1.5.x. Versions of kind >= v1.13.0 use containerd version 1.6.x. To ensure wider test coverage, one test will install kind v1.12.0, while the bulk of the tests will install kind v1.17.0.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -248,6 +248,7 @@ jobs:
         run: |
           export PATH="${KIND_DIR}:${PATH}"
           make e2e-test \
+            KUBERNETES_VERSION=1.23.13 \
             E2E_TESTS=${{ matrix.E2E_TEST }} \
             ERASER_IMG=${ERASER_IMG} \
             MANAGER_IMG=${MANAGER_IMG} \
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -176,6 +176,11 @@ jobs:
           run: make docker-build-collector COLLECTOR_IMG=${COLLECTOR_IMG}
         - name: Build trivy-scanner
           run: make docker-build-trivy-scanner TRIVY_SCANNER_IMG=${SCANNER_IMG}
+        - name: Install kind v0.17.0
```
sozercan, line 179: isn't e2e framework going to override this?

pmengelbert:
e2e framework installs kind if it's not already present

.github/workflows/test.yaml
pmengelbert: e2e framework installs kind if it's not already present, otherwise uses the version present on the machine

pmengelbert:
@sozercan do we need this, or can I close?

PR #510, opened 2022-12-06
fix: metrics fixes

ashnamehrotra:
**What this PR does / why we need it**:
- Moves controller-runtime metrics to same endpoint as otel eraser metrics
- Fixes eraser container metrics to report before writing to named pipe (when scanner was disabled it would return before recording the metrics)

**Which issue(s) this PR fixes** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #509, #502

**Special notes for your reviewer**:

PR #530, opened 2022-12-12
fix: consistently use test namespace for all e2e tests

### LINKED ISSUE #529: fix hardcoded namespace for configmap
["eraser-system" namespace is hardcoded for configmap](https://github.com/Azure/eraser/blob/cdab164c574925c75a3a8fadda71cb1c958cb7c6/controllers/imagecollector/imagecollector_controller.go#L347).

AKS and potentially others do not use the eraser-system namespace, so it shouldn't be hard-coded here.

pmengelbert:
The one exception is the `collector_pipeline` test, which is deployed using the manifest. Absent a way to generate a manifest with an alternative namespace, it's easier to just use the `eraser-system` namespace when deploying from manifest.

Additionally fixes #529 by removing hardcoded reference to the `eraser-system` namespace.

Fixes #529 

This should be merged before #510


test/e2e/tests/collector_pipeline/main_test.go
```diff
@@ -23,7 +23,7 @@ func TestMain(m *testing.M) {
 	// Create KinD Cluster
 	util.Testenv.Setup(
 		envfuncs.CreateKindClusterWithConfig(util.KindClusterName, util.NodeVersion, "../../kind-config.yaml"),
-		envfuncs.CreateNamespace(util.TestNamespace),
+		envfuncs.CreateNamespace(util.EraserNamespace),
```
ashnamehrotra, line 26: we don't need to create the namespace before any of the tests besides metrics and exclusion list, so we should be able to get rid of this in the rest of the tests

ashnamehrotra (changes requested):

test/e2e/tests/collector_pipeline/main_test.go
ashnamehrotra, line 26: we don't need to create the namespace before in the tests besides for metrics and exclusion list so we should be able to delete this in the rest

pmengelbert: This is required now, because it sets `cfg.Namespace()` in the body of the e2e tests, so we need to create one namespace per e2e test

ashnamehrotra (approved):
LGTM

Issue #446, opened 2022-10-28
Use system-wide configmap instead of configuration via cli args


pmengelbert:
**Describe the solution you'd like**
The management of CLI args is getting complex and cluttered. A configmap should be implemented to replace some of the CLI flags for the manager, for example `--collector-pull-policy`, `--collector-arg`, etc. Over time this will get very cluttered, and managing them using the helm chart is getting more and more difficult.

Because options for scanner/eraser/collector containers are specified in code (and thus harder to template), they are currently passed in to cli flags **on the manger**.

A well-structured and well-documented configmap should be implemented. The manager should read this configuration at boot, but it will also have to be configured to respond to updates of the configmap if it is updated by the user while the manager is already deployed.

- Eraser version: v0.4.0

pmengelbert:
A rough draft of what the configmap could look like:

```yaml
# eraser-config.yml
---
eraser:
  image: string
  pullPolicy: corev1.PullPolicy
  runtime: string, "containerd" | "dockershim" | "cri-o" # default "containerd"
  imagelist: string # points to imagelist used for deletion; set to empty when using collector
  profile:
    enable: bool
    port: int32
collector:
  image: string
  pullPolicy: corev1.PullPolicy
  runtime: string, "containerd" | "dockershim" | "cri-o" # default "containerd"
  scanDisabled: bool
  profile:
    enable: bool
    port: int32
scanner:
  image: string
  pullPolicy: corev1.PullPolicy
  profile:
    enable: bool
    port: int32
  cpu:
    request: string
    limit: string
  memory:
    request: string
    limit: string
  scanOptions:
    ignoreUnfixed: bool
    deleteIfScanFailed: bool
    securityChecks:
      - string
      - string
    severities:
      - string
      - string
    vulnerabilityTypes:
      - string
      - string
```

pmengelbert:
A rough proposal of how these values would propagate to their respective components:
- Before creating an ImageJob, the ImageJob controller would check the `eraser-config` (name tbd) configmap in the same namespace as the manager pod.
- The ImageJob controller would use the configmap to set the `image` and `pullPolicy` pieces of the containers it specifies in code.
- The rest of the values will be propagated to the `eraser`, `collector`, and `canner` containers by mounting the configmap into a well-known path for each container.
- The `eraser`, `collector`, `scanner` containers have a configuration struct that matches the schema presented above. It is initialized first with default values.
[...]

PR #555, opened 2022-12-19
chore: Use PodTemplates to support ImageJobs

pmengelbert:
This PR was caused by limits on the size of k8s API requests. By default, `kubectl apply` will add a field, `metadata.annotations.kubectl."kubernetes.io/last-applied-configuration"`, which is a JSON representation of the object. This effectively increases the size of the request.

We ran into size limits when adding a `v1` of the eraser API. Because we want to support a multi-version API, the request to create the CRD for imagejobs was essentially doubled by the addition of the new version.

The component of the CRD that took up the most space was the `ImageJob.Spec` field, a `PodTemplateSpec` type. The generated YAML for this type is extremely long, with a lot of documentation. Being a core kubernetes resource, a `PodTemplateSpec` is very long and complex.

This commit leverages the kubernetes resource `PodTemplate` in order to remove the `ImageJob.Spec` field entirely. Instead of using the `ImageJob` to specify configuration of the `eraser` or `collector` pods, that responsibility is offloaded to a `PodTemplate`.

There is a 1:1 ratio between imagejobs and podtemplates. A given imagejob is expected to have a corresponding podtemplate with the same name, in the same namespace as the controller-manager pod. If no such podtemplate exists, an error is returned and the reconciliation loop spins until the podtemplate is found.

The OwnerReference of the podtemplate is set to its corresponding imagejob. When the imagejob is deleted, the podtemplate will be cleaned up as well.

Finally, I made a change allowing the user to disable the immediate triggering of collector jobs. This was done to avoid situations in the tests where we needed to wait for/delete the initial imagejob in order to test imagelists. Commit 7b49359 should be reverted in a later PR, after #568 has been addressed.


pmengelbert:
reverting to draft status until test issue is worked out

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -163,13 +170,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 				},
 			},
 		}
-		log.Info("Queued first ImageCollector reconcile")
-	}()
+
+		log.Info("...Queued first ImageCollector reconcile")
```
sozercan: why `...` prefix? debugging?

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -154,7 +155,13 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	go func() {
+	delay := *repeatPeriod
+	if *repeatImmediate {
```
sozercan: what is this for? is this different than setting repeat period to 0?

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -233,75 +242,74 @@ func (r *Reconciler) createImageJob(ctx context.Context, req ctrl.Request, argsC
...
+					Name:            "collector",
+					Image:           *collectorImage,
+					ImagePullPolicy: corev1.PullIfNotPresent,
+					Args:            append(collectorArgs, "--scan-disabled="+strconv.FormatBool(scanDisabled)),
+					VolumeMounts: []corev1.VolumeMount{
+						{MountPath: "/run/eraser.sh/shared-data", Name: "shared-data"},
+					},
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							"cpu":    resource.MustParse("7m"),
```
sozercan, line 262: are we hardcoding these?

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -163,13 +170,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-		log.Info("Queued first ImageCollector reconcile")
-	}()
+
+		log.Info("...Queued first ImageCollector reconcile")
+	})
 
 	return nil
 }
 
 //+kubebuilder:rbac:groups="batch",resources=jobs,verbs=get;list;create;delete;watch
```
sozercan: can we remove this or is this used?

ashnamehrotra (approved):

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -359,18 +367,35 @@ func (r *Reconciler) createImageJob(ctx context.Context, req ctrl.Request, argsC
...
+			Namespace: namespace,
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(job, eraserv1alpha1.GroupVersion.WithKind("ImageJob")),
+			},
+		},
+		Template: jobTemplate,
+	}
+	err = r.Create(ctx, &template)
+	if err != nil {
+		log.Info("Could not create collector PodTemplate")
```
sozercan: log.Error?

controllers/imagecollector/imagecollector_controller.go
pmengelbert: These have been hardcoded for some time, the code just moved due to indentation changes, etc.

controllers/imagecollector/imagecollector_controller.go
pmengelbert: We can remove

controllers/imagecollector/imagecollector_controller.go
pmengelbert: :+1:

controllers/imagecollector/imagecollector_controller.go
pmengelbert: The collect-scan-erase process can either begin immediately or after the specified delay. Our default has been to start immediately, with no option to change that behavior. This change makes it possible to do either.

The only reason I added this to this PR is because the collector pods were messing up the tests. Some tests (for example `imagelist_rm_images`) test the behavior of the imagelist, and have to jump through hoops to not conflict with the collector which begins immediately after deployment. example: https://github.com/Azure/eraser/blob/76aa58d70fb1cce6f2c61f466dd4b34dceb470b5/test/e2e/tests/imagelist_rm_images/eraser_test.go#L51

Ideally I'd like to move to a more robust scheduling system (example: allow the user to specify a cron string).

controllers/imagecollector/imagecollector_controller.go
pmengelbert: No reason, will remove

controllers/imagecollector/imagecollector_controller.go
pmengelbert: > what is this for? is this different than setting repeat period to 0?

Sorry, misunderstood your question. Yes, it's different. Here's how it works:
- If repeatPeriod = 30 min and repeatImmediate = true, then we will run one job at startup, then the next job 30 minutes after that
- If repeatPeriod = 30 min and repeatImmediate = false, then we will wait 30 mins after startup to run the first job

On the other hand, setting repeatPeriod to 0 would start immediately and repeat endlessly with no delay

controllers/imagecollector/imagecollector_controller.go
sozercan: i still see this, do we want to remove in this pr?

controllers/imagecollector/imagecollector_controller.go
sozercan: I see, `repeat` seems a bit misleading. perhaps we can name it something like `schedule-immediate`, `queue-immediate` or `erase-immediate` (or anything else you can think of)?

controllers/imagecollector/imagecollector_controller.go
sozercan: is this part of the config we talked about before?

controllers/imagecollector/imagecollector_controller.go
pmengelbert: done

controllers/imagecollector/imagecollector_controller.go
pmengelbert: It should be. I'll leave a link to this comment in the configmap issue #446

controllers/imagecollector/imagecollector_controller.go
pmengelbert: For clarity, you meant the kubebuilder comment? Or the log.Info? Both have been removed.

sozercan: sorry, should have clarified. I meant kubebuilder marker

sozercan (approved):
Thank you! LGTM

Issue #537, opened 2022-12-14
collector pods no longer running in parallel


pmengelbert:
**What steps did you take and what happened:**
The collector pods are now running one-at-a-time instead of all-at-once. This appears to be related to the changes in #515 . We probably need to take action asynchronously when each pod becomes ready.

- Eraser version: v1.0.0-beta.1
- Kubernetes version: (use `kubectl version`):

ashnamehrotra:
fixed in #546

PR #579, opened 2023-01-11
fix: remove accidentally added code

pmengelbert:
The v1 api changes accidentally restored a commit that should have been reverted. This PR removes it again. The code removed here is the cause of #537 .


ashnamehrotra (approved):
LGTM

sozercan (approved):

PR #608, opened 2023-01-30
fix: remove explicit id and group from docker call

pmengelbert:
including the explicit id:group in the `docker run` call breaks permissions in rootless docker setups.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -297,6 +297,5 @@ version-docs:
 	docker run --rm \
 		-v $(shell pwd)/docs:/docs \
 		-w /docs \
-		-u $(shell id -u):$(shell id -g) \
 		node:${NODE_VERSION} \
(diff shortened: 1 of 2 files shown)


ashnamehrotra (approved):

Makefile
```diff
@@ -297,6 +297,5 @@ version-docs:
 	docker run --rm \
 		-v $(shell pwd)/docs:/docs \
 		-w /docs \
-		-u $(shell id -u):$(shell id -g) \
```
sozercan, line 300: is it possible to check if it rootless or otherwise, the files are created as root

pmengelbert: Added a shell script to check for rootless docker and set the uid/gid flags accordingly. Thanks for the heads up on this

pmengelbert:
Will merge once we can rebase on #627

PR #627, opened 2023-02-07
ci: pin golang image to 1.19.4-bullseye

pmengelbert:
There's an issue caused by the 1.19.5-bullseye image: https://github.com/docker-library/golang/issues/452

This will allow CI to succeed while the underlying issue is sorted out

### DIFF
--- a/build/tooling/Dockerfile
+++ b/build/tooling/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.19-bullseye
+FROM golang:1.19.4-bullseye
 
 RUN GO111MODULE=on go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0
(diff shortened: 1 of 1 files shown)


build/tooling/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.19-bullseye
+FROM golang:1.19.4-bullseye
```
sozercan: can we use `golang@sha256:bb9811fad43a7d6fd2173248d8331b2dcf5ac9af20976b1937ecd214c5b8c383` for now? this is Go 1.19.5 without git change

sozercan: let's create an issue so we can track once there's a fix

pmengelbert: Pushed the change, also created this issue to track:

https://github.com/Azure/eraser/issues/630

sozercan (approved):
lgtm pending tests

PR #628, opened 2023-02-07
fix: Ensure windows nodes are excluded by default

pmengelbert:
Helm chart and manifest should define this behavior. Config was not being updated without a call to `Update()`, but we shouldn't be changing the config at runtime (gets out of sync with the configmap).


ashnamehrotra (approved):
tested on windows cluster LGTM after rebase

PR #633, opened 2023-02-14
feat: add cli option for custom runtime path

### LINKED ISSUE #594: Failed to perform collection on k3s cluster  nodes

Attempted to install Eraser on cluster running k3s, and although the manager was able to successfully run, all collectors failed. This failed using both the manifest and the helm installation process, for versions `v0.5.0` and `v0.4.0`, as well as for the manifest installation `v1.0.0-beta.3`.

All version experienced the same error logs on the collector pods:

```
Defaulted container "collector" out of: collector, eraser, trivy-scanner
{"level":"error","ts":1674667563.8705308,"logger":"collector","msg":"failed to get image client","error":"rpc error: code = Unimplemented desc = unknown service runtime.v1.RuntimeService\nrpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService","stacktrace":"main.main\n\t/workspace/pkg/collector/collector.go:60\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
````

Eraser to set up image cleaning

**Environment:**

- Eraser version: `v0.4.0`, `v0.5.0`, `v1.0.0-beta.3`
Client Version: v1.25.2
Kustomize Version: v4.5.7
Server Version: v1.25.4+k3s1

### PR DESCRIPTION
**Pull request by cryptaliagy**
**What this PR does / why we need it**: This PR adds in a CLI flag into the manager component so that it can interpret the runtime option as a literal path.

This will allow non-standard paths (e.g. with the bundled containerd in k3s) for container runtimes to be supported

Fixes #594 

### DIFF
--- a/pkg/eraser/eraser.go
+++ b/pkg/eraser/eraser.go
@@ -26,10 +26,11 @@ import (
 )
 
 var (
-	runtimePtr    = flag.String("runtime", "containerd", "container runtime")
-	imageListPtr  = flag.String("imagelist", "", "name of ImageList")
-	enableProfile = flag.Bool("enable-pprof", false, "enable pprof profiling")
-	profilePort   = flag.Int("pprof-port", 6060, "port for pprof profiling. defaulted to 6060 if unspecified")
+	runtimePtr      = flag.String("runtime", "containerd", "container runtime")
+	isCustomRuntime = flag.Bool("custom-runtime", false, "assume runtime passed is local socket path")
+	imageListPtr    = flag.String("imagelist", "", "name of ImageList")
+	enableProfile   = flag.Bool("enable-pprof", false, "enable pprof profiling")
+	profilePort     = flag.Int("pprof-port", 6060, "port for pprof profiling. defaulted to 6060 if unspecified")
 
 	// Timeout  of connecting to server (default: 5m).
 	timeout  = 5 * time.Minute
@@ -61,6 +62,14 @@ func main() {
 	}
 
 	socketPath, found := util.RuntimeSocketPathMap[*runtimePtr]
+	if *isCustomRuntime {
+		socketPath = *runtimePtr
+
+		// check if socketPath exists
+		if _, err := os.Stat(socketPath); err == nil {
+			found = true
+		}
+	}
 	if !found {
 		log.Error(fmt.Errorf("unsupported runtime"), "runtime", *runtimePtr)
 		os.Exit(generalErr)

### CODE REVIEW

File: pkg/eraser/eraser.go [hunk 1]
```diff
@@ -26,10 +26,11 @@ import (
 )
 
 var (
-	runtimePtr    = flag.String("runtime", "containerd", "container runtime")
-	imageListPtr  = flag.String("imagelist", "", "name of ImageList")
-	enableProfile = flag.Bool("enable-pprof", false, "enable pprof profiling")
-	profilePort   = flag.Int("pprof-port", 6060, "port for pprof profiling. defaulted to 6060 if unspecified")
+	runtimePtr      = flag.String("runtime", "containerd", "container runtime")
+	isCustomRuntime = flag.Bool("custom-runtime", false, "assume runtime passed is local socket path")
```
pmengelbert, line 30: I think splitting the behavior of a configuration value (`runtimePtr`) with a bool might introduce too much cognitive overhead to the end user. They then have to hold two different behaviors in their mind for a single variable.

Overall, I think it's better to allow the user to supply the socket path as a separate variable, for example
```go
flag.String("socket-path", "", "path to the CRI runtime socket")
```
If empty string, use the default location for the provided `runtimePtr`, else use the provided socket address.

pkg/eraser/eraser.go
```diff
@@ -61,6 +62,14 @@ func main() {
 	}
 
 	socketPath, found := util.RuntimeSocketPathMap[*runtimePtr]
+	if *isCustomRuntime {
```
pmengelbert, line 65: See the above comment on `socket-path`.

pkg/eraser/eraser.go
pmengelbert, line 30: The custom socket address also needs to make its way to the collector and scanner pods.

pmengelbert, line 30: One other gotcha is that the configuration system [recently moved from cli args to configmaps](https://github.com/Azure/eraser/pull/581). That means that the end-user no longer has a way to set an arg on an eraser container directly, and it would need to be handled via the configmap and propagated down to the container (through this arg, or through an env var).

pmengelbert (changes requested):
Thank you for the PR! I left a few comments which should help point you in the right direction for completing this.

One hurdle that might block progress is that the default scanner, trivy, [only has partial support for setting the runtime socket](https://github.com/aquasecurity/trivy/pull/3599#issuecomment-1430004789), and the way in which the socket is set is different for each of the three supported runtimes.

I know this is a feature that you want, so If you have any questions I'm happy to provide more information on how we can make progress!

sozercan:
ref https://github.com/aquasecurity/trivy/pull/3599

cryptaliagy:
Abandoning this PR since it is hopelessly out-of-date with the release of 1.0.0, will give it another go using the ConfigMap approach

PR #675, opened 2023-03-24
feat: rename collector pods and eraser container

### LINKED ISSUE #453: Rename `collector` pod and controller
The `collector` pod is no longer an accurate name for what it does, because it collects, scans, and erases. It also creates ambiguity between the `collector` pod vs the container with the same name.

Possible: 
~- `ImageCollector` CR -> `ImageFilter` CR~
~- `ImageCollector` Controller -> `ImageFilter` Controller~
- `collector` pod -> `filter` pod
- `collector` container -> stays `collector` container

**Environment:**
- Eraser version: **v0.4.0**

ashnamehrotra:
- Renames "collector" pods to "eraser" pods and changes "eraser" container name to "remover". If scheduled by ImageCollector, the pods will have label "collector", and if scheduled by ImageList they will have label "remover".
- Fixes helm upgrade tests.
- Fixes configmap update reconfiguration. 

Fixes #453


api/v1alpha1/config/config.go
pmengelbert, line 1: We'll also need this for v1, right?

pmengelbert:
Otherwise LGTM

api/v1alpha1/config/config.go
sozercan: is this user facing or backwards compatible?

pmengelbert: this is user-facing, since this defines the configmap keys

pmengelbert: > We'll also need this for v1, right?

It looks like this file doesn't exist in `api/v1`, so obviously this can't be changed there!

pmengelbert (dismissed):

pmengelbert:
Actually, I noticed that `test/e2e/helm-empty-values.yaml` still says `eraser`

pmengelbert:
Should we also fix the docs?

api/v1alpha1/config/config.go
sozercan: why isn't this in v1?

sozercan: ah nm, this is the alpha configmap change. that makes sense

ashnamehrotra:
helm install with repo fails, converting to draft. Helm test is incorrectly passing since eraser image is not set.

test/e2e/util/utils.go
pmengelbert, line 1: in `HelmDeployLatestEraserRelease`, we will want to install helm using `-f <helm-empy-values>` to be sure we are not using the defaults. The `-f` should come before all of the `--set` arguments.
```
		emptyValuesPath, err := filepath.Abs(filepath.Join(wd, "../../test-data/helm-empty-values.yaml"))
		if err != nil {
			return ctx, err
		}

		// start deployment
		allArgs := []string{providerResourceAbsolutePath, "-f", emptyValuesPath}
```

api/v1alpha1/eraserconfig_types.go
```diff
@@ -135,7 +135,7 @@ type RepoTag struct {
 type Components struct {
 	Collector OptionalContainerConfig `json:"collector,omitempty"`
 	Scanner   OptionalContainerConfig `json:"scanner,omitempty"`
-	Eraser    ContainerConfig         `json:"eraser,omitempty"`
```
pmengelbert: We'll need to revert this back to `Eraser` and `eraser`

api/v1alpha1/config/config.go
```diff
@@ -136,9 +136,9 @@ func Default() *v1alpha1.EraserConfig {
 					Config: &defaultScannerConfig,
 				},
 			},
-			Eraser: v1alpha1.ContainerConfig{
```
pmengelbert: Ditto here (revert to eraser instead of remover)

test/e2e/tests/collector_skip_excluded/main_test.go
```diff
@@ -40,13 +40,15 @@ func TestMain(m *testing.M) {
 		util.HelmDeployLatestEraserRelease(util.TestNamespace,
 			"--set", util.ScannerEnable.Set("false"),
 			"--set", util.CollectorEnable.Set("false"),
+			"--set", util.RemoverImageRepo.Set(removerImage.Repo),
+			"--set", util.RemoverImageTag.Set(removerImage.Tag),
```
pmengelbert, line 45: Set the eraser image instead of the remover image, since the old chart uses v1alpha1.

pmengelbert (changes requested):
Change v1alpha1 to v1alpha2 in most (all?) places, so that the configmap changes will take effect.

config/manager/controller_manager_config.yaml
pmengelbert, line 1: We'll need to change `apiVersion: eraser.sh/v1alpha1` to `v1alpha2` here

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -375,7 +375,7 @@ func (r *Reconciler) createImageJob(ctx context.Context) (ctrl.Result, error) {
 		},
 	}
 
-	job := &eraserv1alpha1.ImageJob{
+	job := &eraserv1alpha2.ImageJob{
```
sozercan: i think this was intentionally left on v1alpha1?

ashnamehrotra:
It looks like the tests were failing because the collector controller was never added to the manager since when HelmDeployLatesetEraserRelease() was called in the tests it would set collector enabled to "false", and in lines [95-99](rename-collector-pods) we don't initially add the collector if that is the case. So any configmap update that later wants to use the collector will not work. The tests are passing locally now, but they aren't working in the CI which could be due to the delay since now it will deploy the pods for HelmDeployLatestEraserRelease() before UpgradeEraserHelm()

pmengelbert:
> It looks like the tests were failing because the collector controller was never added to the manager since when HelmDeployLatesetEraserRelease() was called in the tests it would set collector enabled to "false", and in lines [95-99](rename-collector-pods) we don't initially add the collector if that is the case. So any configmap update that later wants to use the collector will not work. The tests are passing locally now, but they aren't working in the CI which could be due to the delay since now it will deploy the pods for HelmDeployLatestEraserRelease() before UpgradeEraserHelm()

What files are those line numbers for?

I believe disabling the collector and scanner is correct, because otherwise a job will start running immediately. The configmap update won't happen if a job is running, so none of the updates will take effect. In that case, we are testing the older version of eraser (the one in the repo), and not the latest version.

ashnamehrotra:
> 

It is for imagecollector-controller. It never ends up adding the imagecollector-controller to the manager because it reaches this line and returns nil:

```
	collCfg := c.Components.Collector
	if !collCfg.Enabled {
		// don't add controller, but don't throw an error either
		return nil
	}

```

Even though the configmap updates correctly, it doesn't reach the reconcile.

pmengelbert:
> > 
> 
> It is for imagecollector-controller. It never ends up adding the imagecollector-controller to the manager because it reaches this line and returns nil:
> 
> ```
> 	collCfg := c.Components.Collector
> 	if !collCfg.Enabled {
> 		// don't add controller, but don't throw an error either
> 		return nil
> 	}
> ```
> 
> Even though the configmap updates correctly, it doesn't reach the reconcile.

I see. Great work investigating this.

I think what we need to do is if the collector or scanner are enabled after being initially disabled, we need to immediately `os.Exit(1)` (with a log). This will crash the manager, but it will restart immediately and initialize with the new config.

This would need to be done in `startConfigWatch`. Something like this:
```diff
diff --git a/main.go b/main.go
index b6b6de81..10befa50 100644
--- a/main.go
+++ b/main.go
@@ -248,13 +248,19 @@ func startConfigWatch(watcher *inotify.Watcher, eraserOpts *config.Manager, file
 				continue
 			}
 
-			cfg, err := getConfig(filename)
+			oldCfg, err := eraserOpts.Read()
+			if err != nil {
+				setupLog.Error(err, "configuration could not be read", "event", ev, "filename", filename)
+
+			}
+
+			newCfg, err := getConfig(filename)
 			if err != nil {
 				setupLog.Error(err, "configuration is missing or invalid", "event", ev, "filename", filename)
 				continue
 			}
 
-			if err = eraserOpts.Update(cfg); err != nil {
+			if err = eraserOpts.Update(newCfg); err != nil {
 				setupLog.Error(err, "configuration update failed")
 				continue
 			}
@@ -265,6 +271,18 @@ func startConfigWatch(watcher *inotify.Watcher, eraserOpts *config.Manager, file
 				continue
 			}
 
+			type check struct {
+				c bool
+				s bool
+			}
+
+			old := check{c: oldCfg.Components.Collector.Enabled, s: oldCfg.Components.Scanner.Enabled}
+			nnew := check{c: newC.Components.Collector.Enabled, s: newC.Components.Scanner.Enabled}
+			if old != nnew {
+				setupLog.Info("configurations differ in an irreconcileable way", "old", old, "new", nnew)
+				os.Exit(1)
+			}
+
 			setupLog.V(1).Info("new configuration", "manager", newC.Manager, "components", newC.Components)
 		case err := <-watcher.Error:
 			setupLog.Error(err, "file watcher error")
```

@sozercan thoughts on this approach?

main.go
```diff
@@ -265,6 +280,18 @@ func startConfigWatch(watcher *inotify.Watcher, eraserOpts *config.Manager, file
 				continue
 			}
 
+			type check struct {
+				collector bool
+				sscanner  bool
+			}
+
+			old := check{collector: oldCfg.Components.Collector.Enabled, sscanner: oldCfg.Components.Scanner.Enabled}
+			new := check{collector: newC.Components.Collector.Enabled, sscanner: newC.Components.Scanner.Enabled}
```
pmengelbert: `new` is a keyword in golang, so you should use another variable name (this is failing the linter)

.github/workflows/e2e-build.yaml
```diff
@@ -8,6 +8,46 @@ on:
...
+        run: |
+          echo REMOVER_REPO=remover >> $GITHUB_ENV
+          echo REMOVER_TAG=test >> $GITHUB_ENV
+      - name: Build remover
+        run: 'make docker-build-remover OUTPUT_TYPE=type=oci,dest=./${REMOVER_REPO}_${REMOVER_TAG}.tar,name=${REMOVER_REPO}:${REMOVER_TAG}'
+      - name: Upload Build Artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ inputs.bucket-id }}
+          path: remover_test.tar
```
pmengelbert: This should replace `build-eraser`, but you've left both.

.github/workflows/e2e-test.yaml
```diff
@@ -73,29 +73,34 @@ jobs:
...
+          echo REMOVER_REPO=$REMOVER_REPO >> $GITHUB_ENV
           echo ERASER_REPO=$ERASER_REPO >> $GITHUB_ENV
           echo MANAGER_REPO=$MANAGER_REPO >> $GITHUB_ENV
           echo COLLECTOR_REPO=$COLLECTOR_REPO >> $GITHUB_ENV
           echo TRIVY_SCANNER_REPO=$TRIVY_SCANNER_REPO >> $GITHUB_ENV
 
-          echo ERASER_TAG=$ERASER_TAG >> $GITHUB_ENV
+          echo REMOVER_TAG=$REMOVER_TAG >> $GITHUB_ENV
           echo MANAGER_TAG=$MANAGER_TAG >> $GITHUB_ENV
+          echo ERASER_TAG=$ERASER_TAG >> $GITHUB_ENV
```
pmengelbert: Why have both `ERASER` and `REMOVER`?

ashnamehrotra: for the e2e tests, HelmDeployLatestEraserRelease() uses Eraser before we call UpgradeEraserHelm() with Remover

.github/workflows/e2e-build.yaml
ashnamehrotra: we need to build both eraser and remover for the e2e tests (same as below)

charts/eraser/values.yaml
```diff
@@ -69,7 +69,7 @@ runtimeConfig:
         # timeout:
         #   total: 23h
         #   perImage: 1h
-    eraser:
```
sozercan: this shouldn't need to be updated. it will be updated automatically when we do a release

api/v1alpha2/imagejob_types.go
pmengelbert, line 1: This file should be deleted and anything referencing the types in this file should use `v1` instead

api/v1alpha2/imagelist_types.go
pmengelbert, line 1: This file should be deleted. Anything referencing the types in this file should instead reference `v1`

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -455,7 +456,7 @@ func (r *Reconciler) createImageJob(ctx context.Context) (ctrl.Result, error) {
 			Name:      job.GetName(),
 			Namespace: namespace,
 			OwnerReferences: []metav1.OwnerReference{
-				*metav1.NewControllerRef(job, eraserv1alpha1.GroupVersion.WithKind("ImageJob")),
+				*metav1.NewControllerRef(job, eraserv1alpha2.GroupVersion.WithKind("ImageJob")),
```
pmengelbert: Use `v1`

controllers/imagejob/imagejob_controller.go
```diff
@@ -117,7 +117,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 			Type: &corev1.Pod{},
 		},
 		&handler.EnqueueRequestForOwner{
-			OwnerType:    &eraserv1alpha1.ImageJob{},
+			OwnerType:    &eraserv1alpha2.ImageJob{},
```
pmengelbert: `v1`

controllers/imagejob/imagejob_controller.go
```diff
@@ -130,7 +130,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 			Type: &corev1.PodTemplate{},
 		},
 		&handler.EnqueueRequestForOwner{
-			OwnerType:    &eraserv1alpha1.ImageJob{},
+			OwnerType:    &eraserv1alpha2.ImageJob{},
```
pmengelbert: `v1`

controllers/imagelist/imagelist_controller.go
```diff
@@ -331,7 +331,7 @@ func (r *Reconciler) handleImageListEvent(ctx context.Context, imageList *eraser
 				util.ImageJobOwnerLabelKey: ownerLabelValue,
 			},
 			OwnerReferences: []metav1.OwnerReference{
-				*metav1.NewControllerRef(imageList, eraserv1alpha1.GroupVersion.WithKind("ImageList")),
+				*metav1.NewControllerRef(imageList, eraserv1alpha2.GroupVersion.WithKind("ImageList")),
```
pmengelbert: `v1`

controllers/imagelist/imagelist_controller.go
```diff
@@ -370,7 +370,7 @@ func (r *Reconciler) handleImageListEvent(ctx context.Context, imageList *eraser
 			Name:      job.GetName(),
 			Namespace: utils.GetNamespace(),
 			OwnerReferences: []metav1.OwnerReference{
-				*metav1.NewControllerRef(job, eraserv1alpha1.GroupVersion.WithKind("ImageJob")),
+				*metav1.NewControllerRef(job, eraserv1alpha2.GroupVersion.WithKind("ImageJob")),
```
pmengelbert: `v1`

controllers/imagelist/imagelist_controller.go
```diff
@@ -381,7 +381,7 @@ func (r *Reconciler) handleImageListEvent(ctx context.Context, imageList *eraser
 		return reconcile.Result{}, err
 	}
 
-	configMap.ObjectMeta.OwnerReferences = []metav1.OwnerReference{*metav1.NewControllerRef(job, eraserv1alpha1.GroupVersion.WithKind("ImageJob"))}
+	configMap.ObjectMeta.OwnerReferences = []metav1.OwnerReference{*metav1.NewControllerRef(job, eraserv1alpha2.GroupVersion.WithKind("ImageJob"))}
```
pmengelbert: `v1`

pkg/scanners/trivy/helpers.go
```diff
@@ -27,7 +27,7 @@ func loadConfig(filename string) (Config, error) {
 		return cfg, err
 	}
 
-	var eraserConfig eraserv1alpha1.EraserConfig
+	var eraserConfig eraserv1alpha2.EraserConfig
```
pmengelbert: This should be `unversioned`

pkg/scanners/trivy/types.go
```diff
@@ -41,8 +41,8 @@ type (
 	}
 
 	TimeoutConfig struct {
-		Total    eraserv1alpha1.Duration `json:"total,omitempty"`
-		PerImage eraserv1alpha1.Duration `json:"perImage,omitempty"`
+		Total    eraserv1alpha2.Duration `json:"total,omitempty"`
+		PerImage eraserv1alpha2.Duration `json:"perImage,omitempty"`
```
pmengelbert: both `unversioned`

pkg/scanners/trivy/types.go
```diff
@@ -81,8 +81,8 @@ func DefaultConfig() *Config {
 			Severities:     []string{severityCritical, severityHigh, severityMedium, severityLow},
 		},
 		Timeout: TimeoutConfig{
-			Total:    eraserv1alpha1.Duration(time.Hour * 23),
-			PerImage: eraserv1alpha1.Duration(time.Hour),
+			Total:    eraserv1alpha2.Duration(time.Hour * 23),
+			PerImage: eraserv1alpha2.Duration(time.Hour),
```
pmengelbert: both `unversioned`

test/e2e/tests/helm_pull_secret_imagelist/eraser_test.go
```diff
@@ -20,22 +20,22 @@ import (
...
+func TestHelmPullSecretImagelist(t *testing.T) {
 	pullSecretsPropagated := features.New("Image Pull Secrets").
 		Assess("All pods should have the correct pull secret", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			c, err := cfg.NewClient()
 			if err != nil {
 				t.Fatal("Failed to create new client", err)
 			}
 
-			imgList := &eraserv1alpha1.ImageList{
+			imgList := &eraserv1alpha2.ImageList{
```
pmengelbert: `v1`

test/e2e/tests/helm_pull_secret_imagelist/eraser_test.go
```diff
@@ -20,22 +20,22 @@ import (
...
 				ObjectMeta: metav1.ObjectMeta{Name: util.Prune},
-				Spec: eraserv1alpha1.ImageListSpec{
+				Spec: eraserv1alpha2.ImageListSpec{
```
pmengelbert: `v1`

test/e2e/tests/imagelist_alias/eraser_test.go
```diff
@@ -118,9 +118,9 @@ func TestEnsureAliasedImageRemoved(t *testing.T) {
 			return ctx
 		}).
 		Assess("Image deleted when referencing by alias", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
-			imgList := &eraserv1alpha1.ImageList{
+			imgList := &eraserv1alpha2.ImageList{
```
pmengelbert: `v1`

test/e2e/tests/imagelist_alias/eraser_test.go
```diff
@@ -118,9 +118,9 @@ func TestEnsureAliasedImageRemoved(t *testing.T) {
...
 				ObjectMeta: metav1.ObjectMeta{Name: util.Prune},
-				Spec: eraserv1alpha1.ImageListSpec{
+				Spec: eraserv1alpha2.ImageListSpec{
```
pmengelbert: `v1`

test/e2e/tests/imagelist_change/eraser_test.go
```diff
@@ -115,7 +115,7 @@ func TestUpdateImageList(t *testing.T) {
 		}).
 		Assess("Deploy imagelist to remove nginx", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			// deploy imageJob config
-			if err := util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1alpha1_imagelist.yaml"); err != nil {
+			if err := util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1alpha2_imagelist.yaml"); err != nil {
```
pmengelbert: `"eraser_v1_imagelist.yaml"`

test/e2e/tests/imagelist_change/eraser_test.go
```diff
@@ -127,7 +127,7 @@ func TestUpdateImageList(t *testing.T) {
 		}).
 		Assess("Update imagelist to prune rest of images", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			// deploy imageJob config
-			if err := util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1alpha1_imagelist_updated.yaml"); err != nil {
+			if err := util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1alpha2_imagelist_updated.yaml"); err != nil {
```
pmengelbert: `"eraser_v1_imagelist_updated.yaml"`, there should be no imagelist yaml file with `v1alpha2` in the title, as the `v1alpha2/imagelist_types` should be removed (we already have v1).

test/e2e/tests/imagelist_include_nodes/eraser_test.go
```diff
@@ -115,12 +115,12 @@ func TestIncludeNodes(t *testing.T) {
 			}
 
 			// deploy imageJob config
-			if err = util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1alpha1_imagelist.yaml"); err != nil {
+			if err = util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1alpha2_imagelist.yaml"); err != nil {
```
pmengelbert: `"eraser_v1_imagelist.yaml"`

pmengelbert (changes requested):

test/e2e/tests/imagelist_change/eraser_test.go
```diff
@@ -127,7 +127,7 @@ func TestUpdateImageList(t *testing.T) {
 		}).
 		Assess("Update imagelist to prune rest of images", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			// deploy imageJob config
-			if err := util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1alpha2_imagelist_updated.yaml"); err != nil {
+			if err := util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1_imagelist_updated.yaml"); err != nil {
```
pmengelbert: This file needs to exist first

test/e2e/tests/imagelist_include_nodes/eraser_test.go
```diff
@@ -115,7 +115,7 @@ func TestIncludeNodes(t *testing.T) {
 			}
 
 			// deploy imageJob config
-			if err = util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1alpha2_imagelist.yaml"); err != nil {
+			if err = util.DeployEraserConfig(cfg.KubeconfigFile(), cfg.Namespace(), "../../test-data", "eraser_v1_imagelist.yaml"); err != nil {
```
pmengelbert, line 118: file doesn't exist

.github/workflows/e2e-build.yaml
```diff
@@ -8,46 +8,46 @@ on:
...
     timeout-minutes: 10
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@6b3083af2869dc3314a0257a42f4af696cc79ba3
+        uses: step-security/harden-runner@1f99358870fe1c846a3ccba386cc2b2246836776
         with:
           egress-policy: audit
-      - name: Set up Go 1.20
-        uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
+      - name: Set up Go 1.19
```
sozercan: looks like these are getting downgraded

api/v1alpha1/eraserconfig_types.go
```diff
@@ -76,7 +76,7 @@ func (r *Runtime) UnmarshalJSON(b []byte) error {
 }
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+// NOTE: yaml tags are required.  Any new fields you add must have yaml tags for the fields to be serialized.
```
sozercan: i thought we reverted this change

manifest_staging/charts/eraser/templates/imagejobs.eraser.sh-customresourcedefinition.yaml
```diff
@@ -115,3 +115,50 @@ spec:
     storage: false
     subresources:
       status: {}
+  - name: v1alpha2
```
sozercan: why do we have v1alpha2 of imagejob and imagelists?

docs/docs/customization.md
```diff
@@ -72,12 +72,12 @@ take place when:
 
 ### Swapping out components
 
-The collector, scanner, and eraser components can all be swapped out. This
+The collector, scanner, and remover components can all be swapped out. This
```
sozercan, line 75: i think we need to change these in versioned docs?

api/v1alpha1/eraserconfig_types.go
pmengelbert: We did, so this should be reverted here too.

manifest_staging/charts/eraser/templates/imagejobs.eraser.sh-customresourcedefinition.yaml
pmengelbert: We shouldn't

docs/docs/customization.md
pmengelbert: Good catch

test/e2e/test-data/eraser_v1_imagelist_updated.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: eraser.sh/v1alpha1
+apiVersion: eraser.sh/v1alpha2
```
pmengelbert: Should be `v1`

pmengelbert (changes requested):

manifest_staging/charts/eraser/templates/imagejobs.eraser.sh-customresourcedefinition.yaml
pmengelbert: I think re-running codegen and manifest gen should fix this, since those files were removed.

test/e2e/test-data/eraser_v1_imagelist_updated.yaml
sozercan: i think we kept this in `v1alpha1` intentionally to test conversion

docs/docs/customization.md
ashnamehrotra: should I change in both docs/ and version-v1.0.x/ or just version-v1.0.x/?

test/e2e/test-data/eraser_v1_imagelist_updated.yaml
ashnamehrotra: would it be better to revert this back to eraser_v1alpha1_imagelist_updated?

.github/workflows/e2e-build.yaml
```diff
@@ -8,46 +8,46 @@ on:
...
-          echo ERASER_TAG=test >> $GITHUB_ENV
-      - name: Build eraser
-        run: 'make docker-build-eraser OUTPUT_TYPE=type=oci,dest=./${ERASER_REPO}_${ERASER_TAG}.tar,name=${ERASER_REPO}:${ERASER_TAG}'
+          echo REMOVER_REPO=remover >> $GITHUB_ENV
+          echo REMOVER_TAG=test >> $GITHUB_ENV
+      - name: Build remover
+        run: 'make docker-build-remover OUTPUT_TYPE=type=oci,dest=./${REMOVER_REPO}_${REMOVER_TAG}.tar,name=${REMOVER_REPO}:${REMOVER_TAG}'
       - name: Upload Build Artifacts
-        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        uses: actions/upload-artifact@v3
```
sozercan: let's keep the digest here

.github/workflows/e2e-build.yaml
```diff
@@ -8,46 +8,46 @@ on:
...
           restore-keys: |
             ${{ runner.os }}-go-
           path: |
             ~/go/pkg/mod
             ~/.cache/go-build
-      - uses: crazy-max/ghaction-github-runtime@747e98cd97bb9cfed8fe0165c821a83139bfdd67 # v2.1.0
+      - uses: crazy-max/ghaction-github-runtime@v2
       - name: Check out code into the Go module directory
-        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        uses: actions/checkout@v3
```
sozercan: same here

docs/docs/customization.md
sozercan: oh my mistake, `docs` becomes the vNext docs

.github/workflows/e2e-build.yaml
```diff
@@ -8,46 +8,46 @@ on:
...
+        uses: step-security/harden-runner@1f99358870fe1c846a3ccba386cc2b2246836776
         with:
           egress-policy: audit
       - name: Set up Go 1.20
         uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
         with:
           go-version: "1.20"
       - name: Setup buildx instance
-        uses: docker/setup-buildx-action@4b4e9c3e2d4531116a6f8ba8e71fc6e2cb6e6c8c # v2.5.0
+        uses: docker/setup-buildx-action@v2
```
sozercan: let's revert to use the hash

.github/workflows/e2e-build.yaml
```diff
@@ -8,46 +8,46 @@ on:
...
         with:
           use: true
-      - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+      - uses: actions/cache@v3.3.1
```
sozercan: let's revert to use the hash

.github/workflows/e2e-build.yaml
```diff
@@ -8,46 +8,46 @@ on:
...
         with:
           key: ${{ runner.OS }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
           path: |
             ~/go/pkg/mod
             ~/.cache/go-build
-      - uses: crazy-max/ghaction-github-runtime@747e98cd97bb9cfed8fe0165c821a83139bfdd67 # v2.1.0
+      - uses: crazy-max/ghaction-github-runtime@v2
```
sozercan: let's revert to use the hash

test/e2e/test-data/eraser_v1_imagelist_updated.yaml
pmengelbert: This hasn't been resolved yet

main.go
```diff
@@ -248,26 +257,51 @@ func startConfigWatch(watcher *inotify.Watcher, eraserOpts *config.Manager, file
...
+			*newConfig, err = eraserOpts.Read()
 			if err != nil {
 				setupLog.Error(err, "unable to read back new configuration")
 				continue
 			}
 
-			setupLog.V(1).Info("new configuration", "manager", newC.Manager, "components", newC.Components)
+			if needsRestart(oldConfig, newConfig) {
+				setupLog.Info("configurations differ in an irreconcileable way, restarting", "old", oldConfig.Components, "new", newConfig.Components)
+				cancel()
```
pmengelbert, line 288:
```suggestion
                                // restarts the manager
				cancel()
```

pkg/scanners/trivy/helpers.go
```diff
@@ -27,7 +27,7 @@ func loadConfig(filename string) (Config, error) {
 		return cfg, err
 	}
 
-	var eraserConfig eraserv1alpha1.EraserConfig
+	var eraserConfig unversioned.EraserConfig
```
pmengelbert, line 30: This should be addressed in a separate PR, but we will want to manage a versioned schema for the scanner config as well.

test/e2e/tests/collector_disable_scan/main_test.go
```diff
@@ -17,9 +17,9 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	utilruntime.Must(eraserv1alpha1.AddToScheme(scheme.Scheme))
+	utilruntime.Must(eraserv1alpha2.AddToScheme(scheme.Scheme))
```
pmengelbert: I don't know what this line is doing. I'm not sure if it should be `v1alpha1` or `v1alpha2`.

pmengelbert (approved):

PR #681, opened 2023-03-27
ci: build images once

pmengelbert:
A lot of work is duplicated in building the images once per matrix node. This PR builds them only once, uploads them as artifacts, then downloads them for each of the tests in the matrix.

In addition, this PR refactors the workflows to avoid the massive duplication between `upgrade.yaml` and `test.yaml`. This way, they won't get out of sync if the build logic changes.

Finally, we will now also test the upgrade workflow when the upgrade workflow file changes.

- The tests are now a bit faster, taking about a maximum of 12 minutes.
- They will also be less error prone, since building images 30+ times has more room for error
- If the images fail to build, it blocks all of the e2e tests, meaning the queues will be shorter.
- Re-running failed tests will be *much* faster, since the images won't have to be re-built.


.github/workflows/upgrade.yaml
```diff
@@ -3,91 +3,39 @@ on:
   push:
     paths:
       - "manifest_staging/charts/**"
+      - ".github/workflows/upgrade.yaml"
```
ashnamehrotra, line 6: why do we add this under paths?

ashnamehrotra: same for pull_request

pmengelbert: This signals that the upgrade workflow should be run when the `upgrade.yaml` file itself gets updated. Otherwise we would have an untested test (that's fun to say)

ashnamehrotra (approved):
LGTM

PR #701, opened 2023-04-12
feat: use stable trivy as a binary

pmengelbert:
Uses a trivy binary rather than attempting to build from their API. Trivy explicitly makes no guarantees about the stability of their internal API. It's a constantly moving target, and keeping our trivy up to date eats up a significant amount of time.

This is a draft PR to see how it does against the tests.


pmengelbert:
- [ ] Supply config as cli args to trivy
- [ ] Disable all runtimes except containerd
- [ ] Fix unit tests

pmengelbert:
- [ ] We need this https://github.com/aquasecurity/trivy/issues/4044

Dockerfile
```diff
@@ -49,6 +50,39 @@ RUN \
...
+WORKDIR /build
+ENV GOCACHE=/root/gocache
+ENV CGO_ENABLED=0
+RUN \
+    --mount=type=cache,target=${GOCACHE} \
+    --mount=type=cache,target=/go/pkg/mod \
+<<EOF
+set -e
+git init
+git remote add origin ${TRIVY_REPO} || git remote set-url origin ${TRIVY_REPO}
```
sozercan, line 78: we can validate and copy the binary instead of rebuilding

pmengelbert:
I'll want to start from scratch once the Trivy changes are made

PR #726, opened 2023-05-04
fix: Testing CI

pmengelbert:


test/e2e/tests/imagelist_alias/main_test.go
```diff
@@ -18,27 +19,33 @@ import (
 
 func TestMain(m *testing.M) {
 	utilruntime.Must(eraserv1alpha1.AddToScheme(scheme.Scheme))
+	utilruntime.Must(eraserv1.AddToScheme(scheme.Scheme))
```
pmengelbert, line 22: Pretty sure we don't need this (and you can remove the import above as well)

pmengelbert:
This was just for testing

PR #794, opened 2023-07-13
docs: move to CNCF code of conduct

salaxander:
As part of onboarding, we need to move to the CNCF code of conduct


pmengelbert (approved):
LGTM

PR #831, opened 2023-08-14
feat: Add option to config to set runtime socket address

### LINKED ISSUE #647: Make containerd.sock configurable in your helm charts
Microk8s runs containerd in `/var/snap/microk8s/common/run/containerd.sock`.

Eraser right now hardcodes the containerd socket path as a constant:
https://github.com/Azure/eraser/blob/dce0e838ea638e383ef8d38f68b3b726e64e66bd/pkg/utils/utils.go#L37

Even without Microk8s, this would reduce the usability of eraser in other environments where the socket is installed somewhere else.

This results in Eraser Collector failing in Microk8s:
```bash
{"level":"error","ts":1677250961.0918322,"logger":"eraser","msg":"failed to get image client","error":"rpc error: code = Unimplemented desc = unknown service runtime.v1.RuntimeService\nrpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService","stacktrace":"main.main\n\t/workspace/pkg/eraser/eraser.go:71\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
```
An easy solution is to expose the Containerd Socket parameter in your Helm chart, and injecting that into Go as a ConfigMap.

**Environment:**

- Eraser version: `kubectl apply -f https://raw.githubusercontent.com/Azure/eraser/v1.0.0/deploy/eraser.yaml`
- Kubernetes version: (use `kubectl version`): Microk8s `v1.26.1`

pmengelbert:

Fixes #647


api/v1alpha3/config/config.go
sozercan, line 1: why are we adding a new api version

pmengelbert: I wasn't sure if it was merited here or not. I don't mind adding the changes to `v1alpha2` instead

sozercan: that should be fine as long as we default if users don't specify. wdyt?

pmengelbert: `v1alpha3` defaults to containerd and the default containerd socket address. If it's a version prior to `v1alpha3` and something other than containerd is specified, then it'll use the default socket address for that runtime.

pmengelbert:
Superseded by #930

PR #922, opened 2023-11-30
feat: add v1alpha3 api with RuntimeSpec configuration

ashnamehrotra:
Part 1 of adding RuntimeSpec configuration. Adds v1alpha3 api with RuntimeSpec field and relevant conversion functions.

Fixes #


api/v1alpha3/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+  - MEDIUM
+  - LOW
+`
+
+type Manager struct {
+	mtx sync.Mutex
+	cfg *v1alpha3.EraserConfig
+}
+
+func (m *Manager) Read() (v1alpha3.EraserConfig, error) {
```
pmengelbert: I believe you can remove these `Read` and `Update` methods and the `NewManager` function. Only the `unversioned` package needs those. For example, the corresponding functions from the `v1alpha2/config` package are not actually called anywhere.

I think the functions aren't even needed for the conversion generation either. You can remove these, and the corresponding ones in `v1alpha1` and `v1alpha2` as well.

api/v1alpha3/eraserconfig_types.go
```diff
@@ -0,0 +1,209 @@
...
+
+	ContainerdPath = "/run/containerd/containerd.sock"
+	DockerPath     = "/run/dockershim.sock"
+	CrioPath       = "/run/crio/crio.sock"
+)
+
+func ConvertRuntimeToRuntimeSpec(r Runtime) (RuntimeSpec, error) {
+	var rs RuntimeSpec
+
+	switch rt := Runtime(r); rt {
```
pmengelbert: The cast to the `Runtime` type is unnecessary because `r` is already a `Runtime` (it was wrong in my draft PR too). You can just `switch r`.

api/v1alpha3/custom_conversions.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+//nolint:revive
+func Convert_unversioned_ManagerConfig_To_v1alpha3_ManagerConfig(in *unversioned.ManagerConfig, out *ManagerConfig, s conversion.Scope) error {
+	return autoConvert_unversioned_ManagerConfig_To_v1alpha3_ManagerConfig(in, out, s)
+}
+
+// TODO: change this to use unversioned.RuntimeSpec when unversioned is updated
+//
+//nolint:revive
```
pmengelbert, line 33:
```suggestion
//nolint:revive
```

api/v1alpha3/custom_conversions.go
```diff
@@ -0,0 +1,43 @@
...
+func Convert_v1alpha3_RuntimeSpec_To_unversioned_Runtime(in *RuntimeSpec, out *unversioned.Runtime, s conversion.Scope) error {
+	return autoConvert_v1alpha3_RuntimeSpec_To_unversioned_Runtime(in, out, s)
+}
+
+//nolint:revive
+func Convert_unversioned_ManagerConfig_To_v1alpha3_ManagerConfig(in *unversioned.ManagerConfig, out *ManagerConfig, s conversion.Scope) error {
+	return autoConvert_unversioned_ManagerConfig_To_v1alpha3_ManagerConfig(in, out, s)
+}
+
+// TODO: change this to use unversioned.RuntimeSpec when unversioned is updated
```
pmengelbert: A good practice is to `git grep TODO` before submitting a PR. Any TODOs should be either implemented or filed as issues.

api/v1alpha3/eraserconfig_types.go
```diff
@@ -0,0 +1,209 @@
...
+	RuntimeContainerd Runtime = "containerd"
+	RuntimeDockerShim Runtime = "dockershim"
+	RuntimeCrio       Runtime = "crio"
+
+	ContainerdPath = "/run/containerd/containerd.sock"
+	DockerPath     = "/run/dockershim.sock"
+	CrioPath       = "/run/crio/crio.sock"
+)
+
+func ConvertRuntimeToRuntimeSpec(r Runtime) (RuntimeSpec, error) {
```
pmengelbert, line 49: I'd like to see a unit test for this.

api/v1alpha3/eraserconfig_types.go
```diff
@@ -0,0 +1,209 @@
...
+	case RuntimeCrio:
+		rs = RuntimeSpec{Name: RuntimeCrio, Address: RuntimeAddress(fmt.Sprintf("unix://%s", CrioPath))}
+	default:
+		return rs, fmt.Errorf("invalid runtime: valid names are %s, %s, %s", RuntimeContainerd, RuntimeDockerShim, RuntimeCrio)
+	}
+
+	return rs, nil
+}
+
+func (td *Duration) UnmarshalJSON(b []byte) error {
```
pmengelbert, line 66: I'd also like to have a unit test to prevent this from regressing. The unmarshaling is important because it catches errors, and we need to have tests to prove it fails when it is supposed to and doesn't when it's not.

api/v1alpha3/custom_conversions.go
```diff
@@ -0,0 +1,43 @@
...
+//
+//nolint:revive
+func autoConvert_unversioned_Runtime_To_v1alpha3_RuntimeSpec(in *unversioned.Runtime, out *RuntimeSpec, _ conversion.Scope) error {
```
pmengelbert: This shouldn't be named `autoConvert...`, since it's manually written, right?

pmengelbert (changes requested):
This PR isn't quite complete. We need to replace the `unversioned.EraserConfig` with the `v1alpha3` version. The `unversioned` should always be the latest stable version. Since none of our `EraserConfig` versions are stable, it should always just be the latest.

This will have effects on some of the code here, since it won't be necessary to have conversion functions between `v1alpha3` and `unversioned`. But you will need converters for 1 -> unversioned and 2 -> unversioned.

pmengelbert:
On second thought, it's best to keep the `unversioned` api at `v1alpha2`, and to switch it to `v1alpha3` in the future when it's properly wired up.

api/v1alpha3/eraserconfig_types.go
```diff
@@ -0,0 +1,213 @@
...
+
+	ContainerdPath = "/run/containerd/containerd.sock"
+	DockerPath     = "/run/dockershim.sock"
+	CrioPath       = "/run/crio/crio.sock"
+)
+
+func ConvertRuntimeToRuntimeSpec(r Runtime) (RuntimeSpec, error) {
+	var rs RuntimeSpec
+
+	switch rt := r; rt {
```
pmengelbert: Just `switch r` here

api/v1alpha3/eraserconfig_types.go
```diff
@@ -0,0 +1,214 @@
...
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type (
+	Duration       time.Duration
+	Runtime        string
+	RuntimeAddress string
```
pmengelbert: Why the newtype for `RuntimeAddress`? Why not just use a string?

api/v1alpha3/runtime_spec_test.go
```diff
@@ -0,0 +1,110 @@
...
+func TestUnmarshalJSON(t *testing.T) {
+	type testCase struct {
+		input     []byte
+		expected  RuntimeSpec
+		shouldErr bool
+	}
+
+	tests := map[string]testCase{
+		"ValidContainerd": {
+			input:     []byte(`{"Name": "containerd", "Address": "unix:///run/containerd/containerd.sock"}`),
```
pmengelbert: The keys should all be lowercase, i.e. `name` and not `Name`.  If the test is passing as-is, it is a false positive and we need to find out why.

pmengelbert: And by keys I mean the keys in the JSON byte string.

pkg/scanners/trivy/helpers.go
```diff
@@ -12,12 +12,14 @@ func loadConfig(filename string) (Config, error) {
 
 	b, err := os.ReadFile(filename)
 	if err != nil {
+		log.Info("LOADCONFIG unable to read file")
```
pmengelbert: Maybe `log.Error` here, and remove the `LOADCONFIG` -- the logger should give us more info on where the error is occurring.

api/v1alpha3/runtime_spec_test.go
```diff
@@ -0,0 +1,110 @@
...
+			input:     []byte(`{"Name": "containerd", "Address": "http://invalid"}`),
+			expected:  RuntimeSpec{},
+			shouldErr: true,
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			var rs RuntimeSpec
+			err := rs.UnmarshalJSON(test.input)
```
pmengelbert: I think we shouldn't call `UnmarshalJSON` directly but rather just `err := json.Unmarshal(test.input, &rs)`. That's because `json.Unmarshal` will ultimately call `rs.UnmarshalJSON`. That's how it ends up getting used in our code, so that's what we should test.

api/v1alpha3/runtime_spec_test.go
ashnamehrotra: changed keys to lowercase! it says here it accepts case-insensitive match for the keys: https://pkg.go.dev/encoding/json#Unmarshal

api/v1alpha3/eraserconfig_types.go
ashnamehrotra: changed to string only Runtime has its own type

pmengelbert (approved):
LGTM, thanks!

PR #941, opened 2024-01-10
chore: update to go 1.21 and trivy 0.48.3

sozercan:
- update go to 1.21
- update trivy to 0.48.2
- move base images inline so dependabot will pin and bump

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,7 @@ REMOVER_TAG ?= ${VERSION}
 # Image URL to use all building/pushing image targets
 TRIVY_SCANNER_REPO ?= ghcr.io/eraser-dev/eraser-trivy-scanner
 TRIVY_SCANNER_IMG ?= ${TRIVY_SCANNER_REPO}:${TRIVY_SCANNER_TAG}
-TRIVY_BINARY_REPO ?= ghcr.io/aquasecurity/trivy
-TRIVY_BINARY_TAG ?= 0.45.1
-TRIVY_BINARY_IMG ?= ${TRIVY_BINARY_REPO}:${TRIVY_BINARY_TAG}
+TRIVY_BINARY_VERSION ?= 0.48.3
 MANAGER_REPO ?= ghcr.io/eraser-dev/eraser-manager
 MANAGER_IMG ?= ${MANAGER_REPO}:${MANAGER_TAG}
 REMOVER_REPO ?= ghcr.io/eraser-dev/remover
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ REMOVER_TAG ?= ${VERSION}
 TRIVY_SCANNER_REPO ?= ghcr.io/eraser-dev/eraser-trivy-scanner
 TRIVY_SCANNER_IMG ?= ${TRIVY_SCANNER_REPO}:${TRIVY_SCANNER_TAG}
 TRIVY_BINARY_REPO ?= ghcr.io/aquasecurity/trivy
-TRIVY_BINARY_TAG ?= 0.45.1
+TRIVY_BINARY_TAG ?= 0.48.1
 TRIVY_BINARY_IMG ?= ${TRIVY_BINARY_REPO}:${TRIVY_BINARY_TAG}
```
ashnamehrotra, line 13: I think we can get rid of the TRIVY_BINARY_IMG if the arg is removed from the Dockerfile?

sozercan: good catch, removed

ashnamehrotra:
we might need to change this value too since it needs TRIVY_BINARY_TAG: `TRIVY_SCANNER_LDFLAGS ?= $(ERASER_LDFLAGS) -X 'main.trivyVersion=v$(TRIVY_BINARY_TAG)'`

ashnamehrotra:
one more thing, we should remove trivyVersion from the trivy.go logs (line 69) since we aren't passing it through the linker anymore

sozercan:
@ashnamehrotra I didn't realized we had that. That seems valuable. Reverted some of the changes and created `TRIVY_BINARY_VERSION` variable

ashnamehrotra (approved):
LGTM

sozercan:
keeping trivy arg in dockerfile for now. if dependabot updates this, then makefile (version is used in ldflags) and dockerfile will be out of sync

Makefile
```diff
@@ -228,7 +226,6 @@ docker-build-trivy-scanner: ## Build docker image for trivy-scanner image.
 	docker buildx build \
 		$(_CACHE_FROM) $(_CACHE_TO) \
 		$(_ATTESTATIONS) \
-		--build-arg TRIVY_BINARY_IMG="$(TRIVY_BINARY_IMG)" \
```
pmengelbert: Why are we removing this?

sozercan: Reverted this change. I was a bit confused since we define in both Makefile and Dockerfile, but I see that Dockerfile gets overwritten by this (added a comment in Dockerfile). It would be good to have a single source of truth in the future.

pmengelbert (approved):
LGTM

PR #948, opened 2024-01-12
chore: group dependabot PRs

### LINKED ISSUE #878: group dependabot prs
https://github.com/eraser-dev/eraser/blob/main/.github/dependabot.yml

group where it makes sense

for example:
- gh actions group
- npm group

**Environment:**

- Eraser version:

ashnamehrotra:

Fixes #878


.github/dependabot.yml
```diff
@@ -26,6 +34,10 @@ updates:
         update-types:
         - "version-update:semver-major"
         - "version-update:semver-minor"
+    groups:
+      all:
+        patterns:
+        - "*"
```
sozercan: we shouldn't do * here, but only for things we want to group (for example k8s.io) but I remember CRI had to be specific version to not break stuff

.github/dependabot.yml
```diff
@@ -7,13 +7,21 @@ updates:
       interval: "weekly"
     commit-message:
       prefix: "chore"
+    groups:
+      all:
+        patterns:
+        - "*"
```
sozercan: might want to group specific things here too, like docuausuarus ones instead of *

.github/dependabot.yml
```diff
@@ -26,6 +34,10 @@ updates:
         update-types:
         - "version-update:semver-major"
         - "version-update:semver-minor"
+    groups:
+      k8s:
+        patterns:
+        - "k8s.io/kubernetes"
```
sozercan: this is a single dependency. we can do k8s.io/* but we'll have to exclude cri. looks like we can have exclusions: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

ashnamehrotra: ah got it! will change it to exclude cri

sozercan (approved):
Thanks! LGTM

PR #957, opened 2024-01-18
chore: update download and upload artifact to latest

ashnamehrotra:
Updates download and upload artifact to latest. Upload and Download artifact have to be set to the same major version. Upload artifact cut a new release following v4.1.0 that adds ability to overwrite an artifact (https://github.com/actions/upload-artifact/releases) which was throwing an error in PR #954. v4 also doesn't allow mutable artifacts so we have to give each (collector, remover, manager, scanner) a unique name.

Fixes #


ashnamehrotra:
overwrite:true feature was flakey in the first run (https://github.com/eraser-dev/eraser/actions/runs/7577112778/attempts/1?pr=957) but the tests passed in the second

sozercan (approved):
Thanks! LGTM

PR #965, opened 2024-01-25
feat: fix manager memory usage

### LINKED ISSUE #798: fix manager memory usage
Modify ImageJob handleNewJob() implementation as the manager currently has high memory usage due to storing the node list on every imagejob run - https://github.com/Azure/eraser/blob/main/controllers/imagejob/imagejob_controller.go#L288. 

**Environment:**

- Eraser version:

ashnamehrotra:
Utilizes `cache.BuilderWithOptions` in the manager to free up space from watching unnecessary resources and changes ImageJob controller to use `nodeNames` to retrieve each node rather than storing all nodes in a list.

Fixes #798 

Eraser load test results for cluster with 500 vuln images, 50 nodes, cpu request unset, and high usage daemonset deployed 

Before changes:
MAX CPU Manager: 973.889567 m
MAX MEM Manager: 351.07 Mi

After changes:
MAX CPU Manager: 441.082336 m
MAX MEM Manager: 32.94 Mi


sozercan:
@ashnamehrotra this is awesome! looks like e2e has a conflict due to artifact names and please sign the DCO

main.go
```diff
@@ -104,6 +108,26 @@ func main() {
 		Port:                   9443,
 		HealthProbeBindAddress: ":8081",
 		LeaderElection:         false,
+		NewCache: cache.BuilderWithOptions(cache.Options{
+			SelectorsByObject: cache.SelectorsByObject{
+				// to watch eraser pods
+				&corev1.Pod{}: {
+					Field: fields.OneTermEqualSelector("metadata.namespace", utils.GetNamespace()),
```
enj, line 115: I _think_ this is still making a cluster scoped watch and having API server filter them down.  You could confirm that one way or the other by reducing the RBAC of the eraser components to use role bindings to properly scope the access.

controllers/imagejob/imagejob_controller.go
```diff
@@ -325,6 +325,13 @@ func (r *Reconciler) handleNewJob(ctx context.Context, imageJob *eraserv1.ImageJ
 		return err
 	}
 
+	var nodeNames []string
```
enj: I would try checking memory usage without this change (unclear to me if it is actually helping).

ashnamehrotra: yes thats correct! I removed this

main.go
ashnamehrotra: The RBAC is already scoped to the namespace for the manager role, but removing the field to filter by namespace increases the memory again. I think that could be because for the manager we have ClusterRole instead of Role, but we can't change that since we need to watch ImageJobs/ImageLists in default and eraser pods in eraser-system. 

It was the same problem testing with MultiNamespacedCacheBuilder and keeping only the eraser-system and default namespaces, it didn't have an affect on memory consumption.

enj: You should create separate cluster roles/roles/cluster roles bindings/role bindings to express the exact permissions of each component.  There is no requirement that everything be done via a single cluster role/cluster role binding.

enj: https://youtu.be/Nw1ymxcLIDI is a good resource to understand RBAC fundamentals.

ashnamehrotra: thank you this was really helpful! I changed the role to create a separate Role and ClusterRole

manifest_staging/charts/eraser/templates/eraser-manager-role-role.yaml
enj, line 1: The cluster role in `manifest_staging/charts/eraser/templates/eraser-imagejob-pods-cluster-role-clusterrole.yaml` is empty, should and the associated binding it be deleted?

manifest_staging/charts/eraser/templates/eraser-manager-role-clusterrole.yaml
```diff
@@ -28,29 +16,6 @@ rules:
...
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
 - apiGroups:
   - eraser.sh
```
enj, line 20: For the eraser resources, does your controller need access to `create` and similar permissions on the resource itself (i.e. not `/status`)?

config/rbac/role_binding.yaml
```diff
@@ -1,10 +1,11 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: RoleBinding
 metadata:
   name: manager-rolebinding
+  namespace: system
```
enj, line 5: Should this be `eraser-system`?

enj (commented):
Minor comments, RBAC LGTM.

manifest_staging/charts/eraser/templates/eraser-manager-role-role.yaml
ashnamehrotra: yup we should only need the service account

config/rbac/role_binding.yaml
ashnamehrotra: by default the roles are in "system" and during deployment it changes to "eraser-system" since the namespace is configurable

sozercan: kubebuilder/kustomize automatically prefixes these
https://github.com/eraser-dev/eraser/blob/e6c15ca505f07c0e968031f5b739c3a762735c32/config/default/kustomization.yaml#L9

controllers/imagejob/imagejob_controller.go
```diff
@@ -189,9 +189,9 @@ func checkNodeFitness(pod *corev1.Pod, node *corev1.Node) bool {
 }
 
 //+kubebuilder:rbac:groups=eraser.sh,resources=imagejobs,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=podtemplates,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",namespace="system",resources=podtemplates,verbs=get;list;watch;create;update;patch;delete
```
sozercan, line 192: please make sure system doesn't break since kubebuilder adds that prefix in generation but not sure if it applies here

ashnamehrotra: this only generates the files under rbac/ which all use system. Then in manifest staging/ it is replaced to eraser-system through the templates.

manifest_staging/charts/eraser/templates/eraser-manager-role-clusterrole.yaml
ashnamehrotra: It does for ImageJob, but not for ImageList. Removed create/delete/update/patch permissions for ImageList. Removed update/patch permissions for ImageJob.

sozercan (approved):
Great job! LGTM

PR #991, opened 2024-02-26
feat: add option to add additional custom pod labels

patrick-vonsteht:
This PR adds the option to configure additional custom pod labels for all pods created by the controller deployment.

This makes it possible to use the controller in environments that require the use of certain labels on all pods, e.g. for governance reasons. One concrete example would be the requirement to label each pod with the cost center that pays for it.

The PR adds an additionalPodLabels parameter, which is a pattern that is used by many helm charts to solve this issue.

To configure the additional labels, you need to do the following:
(A) For the helm chart deployment, add to your values.yaml:
   ```
   deploy:
     additionalPodLabels:
       foo: bar
   runtimeConfig:
     manager:
       additionalPodLabels:
         foo: bar
   ```

(B) For the manifest deployment, edit your manifest:
   * Add the label to the eraser-controller-manager pod template
   * Add to the contents of the eraser-manager-config config map:
   ```
   manager:
     additionalPodLabels:
       foo: bar
   ```

None. This PR proposes this feature.

Thank you for looking at this. I'm looking forward to your feedback and hope we can add this feature to the controller.


pmengelbert:
@patrick-vonsteht Thanks for the PR! I'll take a look today

api/v1alpha1/eraserconfig_types.go
```diff
@@ -91,15 +91,16 @@ type ContainerConfig struct {
...
+	Runtime             Runtime           `json:"runtime,omitempty"`
+	OTLPEndpoint        string            `json:"otlpEndpoint,omitempty"`
+	LogLevel            string            `json:"logLevel,omitempty"`
+	Scheduling          ScheduleConfig    `json:"scheduling,omitempty"`
+	Profile             ProfileConfig     `json:"profile,omitempty"`
+	ImageJob            ImageJobConfig    `json:"imageJob,omitempty"`
+	PullSecrets         []string          `json:"pullSecrets,omitempty"`
+	NodeFilter          NodeFilterConfig  `json:"nodeFilter,omitempty"`
+	PriorityClassName   string            `json:"priorityClassName,omitempty"`
+	AdditionalPodLabels map[string]string `json:"additionalPodLabels,omitempty"`
```
pmengelbert: Please remove this change, as well as the one to `api/v1alpha2/eraserconfig_types.go` (and any autogenerated files that follow from those). We'd like to leave those unchanged. `unversioned` should then be identical to `v1alpha3`.

pmengelbert (changes requested):
You will also need to run `make manifests`. This will generate the staging manifest and helm chart under `manifest_staging`.

api/v1alpha1/eraserconfig_types.go
patrick-vonsteht: I have also committed the manifest_staging directory now.

Also, I have removed the changes to api/v1alpha1 and api/v1alpha2 and ran the autogeneration again.

Was it correct that I also removed my changes to api/v1alpha1/config/config.go and api/v1alpha2/config/config.go? You didn't mention these explicitly, but I assume you want to keep everything stable in api/v1alpha1 and api/v1alpha2?

pmengelbert: Yes, that's correct. I'll approve the CI run now

ashnamehrotra:
@patrick-vonsteht the api changes lgtm, thanks! we will also need to change the controller code to set the labels if they are present in the config (https://github.com/eraser-dev/eraser/blob/main/controllers/imagecollector/imagecollector_controller.go#L326) before creating the ImageJob.

patrick-vonsteht:
@ashnamehrotra thank you for the hint, I'll take a look at that.

patrick-vonsteht:
@ashnamehrotra I have extended the code to add the additional labels also to the ImageJob PodTemplateSpec.

patrick-vonsteht:
Looks like all the tests are successful, except the vulnerability scan. But the vulnerabilities are unrelated to my changes, because I didn't change any dependencies. How should I proceed with this?

ashnamehrotra:
> Looks like all the tests are successful, except the vulnerability scan. But the vulnerabilities are unrelated to my changes, because I didn't change any dependencies. How should I proceed with this?

Since that is a vulnerability in the trivy binary we can ignore that.

controllers/imagecollector/imagecollector_controller.go
```diff
@@ -323,7 +323,15 @@ func (r *Reconciler) createImageJob(ctx context.Context) (ctrl.Result, error) {
...
 
+	labels := map[string]string{}
+	for k, v := range mgrCfg.AdditionalPodLabels {
+		labels[k] = v
+	}
+
 	jobTemplate := corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: labels,
+		},
```
ashnamehrotra: since we are also setting this in the imagejob_controller we can get rid of the functionality here

controllers/imagelist/imagelist_controller.go
```diff
@@ -291,7 +291,15 @@ func (r *Reconciler) handleImageListEvent(ctx context.Context, imageList *eraser
...
 
+	labels := map[string]string{}
+	for k, v := range eraserConfig.Manager.AdditionalPodLabels {
+		labels[k] = v
+	}
+
 	jobTemplate := corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: labels,
+		},
```
ashnamehrotra: same here, we don't need to set it here since its done in the imagejob_controller already

manifest_staging/charts/eraser/values.yaml
```diff
@@ -97,6 +98,7 @@ deploy:
     tag: "v1.4.0-beta.0"
   additionalArgs: []
   priorityClassName: ""
+  additionalPodLabels: {}
```
ashnamehrotra, line 101: why do we need additionalPodLabels here?

ashnamehrotra:
> Looks like all the tests are successful, except the vulnerability scan. But the vulnerabilities are unrelated to my changes, because I didn't change any dependencies. How should I proceed with this?

I can put up a separate PR to update that vulnerability.

manifest_staging/charts/eraser/values.yaml
patrick-vonsteht: The goal is to be able to set additionalPodLabels on every pod. So we have to handle two parts:
1.) The pod labels for the eraser-controller-manager pod. We define these in deploy.additionalPodLabels, because they are used during the deployment of the controller.
2.) The pod labels for the pods that the controller creates at runtime. We define these in runtimeConfig.manager.additionalPodLabels, because they are used during runtime of the controller.

controllers/imagecollector/imagecollector_controller.go
patrick-vonsteht: I have removed it.

controllers/imagelist/imagelist_controller.go
patrick-vonsteht: I have removed it.

sozercan:
fixable vulns should already be addressed with #1002. trivy didn't had a release at the time with a fix so we can't fix that yet. vuln check is meant to be informative, and not blocking so we can ignore that for this pr specifically.

manifest_staging/charts/eraser/values.yaml
ashnamehrotra: got it, thanks for clarifying!

ashnamehrotra (approved):
LGTM, thank you!

pmengelbert (approved):
LGTM, thanks!

PR #1006, opened 2024-03-27
fix: set ignoreUnfixed false

sozercan:

This is a behavior change so we should document this in the release notes.

Changes default behavior to remove vulnerable images regardless of whether they have a fix or not

Fixes #


ashnamehrotra (approved):
LGTM

Issue #1150, opened 2025-05-07
[REQ] Path lookup for trivy if /trivy is not there


ashnamehrotra:
### What kind of request is this?

None

### What is your request or suggestion?

Eraser should do a $PATH lookup for trivy if /trivy is not there. Related to #1149

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

PR #1154, opened 2025-06-12
fix: trivy binary check fallback

### LINKED ISSUE #1149: [BUG] trivy-scanner pod doesn't error when trivy binary cannot be found
### Version of Eraser

v1.4.0

### Expected Behavior

When the trivy binary cannot be found at `/trivy` the pod running trivy-scanner logs an error message that the binary cannot be found, but the pod doesn't exit or error in a way that surfaces the failure to find the binary

### Actual Behavior

_No response_

### Steps To Reproduce

_No response_

### Are you willing to submit PRs to contribute to this bug fix?

- [ ] Yes, I am willing to implement it.

sozercan:
This isn't something that affects the official builds from GHCR since container builds always copies trivy to `/trivy`. 

If the project is rebuilt and the binary is not placed in root dir, this causes issues since eraser expects trivy to be in root. This PR enables fallback to `PATH`. If not found in `PATH`, it will provide an error.

Fixes #1149 #1150


ashnamehrotra (approved):
LGTM thanks!

.golangci.yaml
```diff
@@ -20,7 +20,6 @@ linters:
   disable-all: true
   enable:
     - errcheck
-    - exportloopref
```
pmengelbert, line 23: Why is this check being disabled?

build/tooling/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.23-bullseye@sha256:80eb3147ef40b58d527d9c2e634b1a79a750aee09de6f973844db38b33f0550b
```
pmengelbert, line 1: Can you put the version bumps in a separate PR?

sozercan: unfortunately, setup-env tests were failing because it's now requiring an updated Go. that's the reason for this bump, since it's a dependency for tests

.golangci.yaml
sozercan: `ERRO [linters_context] exportloopref: This linter is fully inactivated: it will not produce any reports. `

this makes lint fail since linter is inactivated

sozercan: replaced with `copyloopvar`

pkg/scanners/trivy/types.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/eraser-dev/eraser/pkg/utils"
 )
 
+// currentExecutingLookPath is a variable that points to exec.LookPath by default,
+// but can be overridden for testing purposes.
+var currentExecutingLookPath = exec.LookPath
```
pmengelbert: It's a little weird to have this as a global variable here. Why not just use `exec.LookPath` in the `findTrivyExecutable` function?

Instead of overriding the lookup function for testing, why not create a `t.TempDir()`, create some executable files in it, and maniuplate `PATH` by using `os.Setenv`?

pkg/scanners/trivy/types.go
```diff
@@ -186,13 +191,13 @@ func (s *ImageScanner) Scan(img unversioned.Image) (ScanStatus, error) {
 		stderr := new(bytes.Buffer)
 
 		cliArgs := s.config.cliArgs(refs[i])
-		cmd := exec.Command(trivyCommandName, cliArgs...)
+		cmd := exec.Command(s.trivyPath, cliArgs...) // nolint:gosec // G204: Subprocess launched with variable
```
pmengelbert, line 190: Can you make the `initScanner` function create a symlink at `/trivy` to the path where trivy was found? That way we don't have to disable a security lint.

pmengelbert (changes requested):
Few comments

pkg/scanners/trivy/types.go
sozercan: good point, this might be why we used `/trivy` path instead of relying on `PATH`. However, will this break read only root file system pod security?

pmengelbert: On second thought, we can change `/trivy` to `trivy` in the call to `exec.Command` and it will be located in the `PATH` automatically

Thanks to Mo for the fresh perspective :)

PR #1238, opened 2026-08-26
fix: give the worker handoff endpoints a Windows path

charleswool:
## What

Gives the four worker handoff endpoints a Windows path. They were POSIX strings with no build tag:

```go
ScanErasePath = "/run/eraser.sh/shared-data/scanErase"
```

## Why

The manager mounts the shared-data `emptyDir` at a Windows path on a Windows node, but the worker binary compiled for `windows/amd64` still looked for its peer under `/run/eraser.sh`. The pod starts, and the collector then fails to bind its endpoint.

This is the gap between *"a worker pod reaches Running on a Windows node"* and *"unused image cleanup works on a Windows node"*. #1229 made the CRI reachable and #1231 made the handoff itself work; this is the last piece of the worker's own view of the filesystem.

## How

The constants move into the platform files, next to `CRIPath`, which already splits for exactly the same reason:

| | `platform_unix.go` | `platform_windows.go` |
|---|---|---|
| `ScanErasePath` | `/run/eraser.sh/shared-data/scanErase` | `C:\run\eraser.sh\shared-data\scanErase` |
| `CollectScanPath` | `…/collectScan` | `…\collectScan` |
| `EraseCompleteCollectPath` | `…/eraseCompleteCollect` | `…\eraseCompleteCollect` |
| `EraseCompleteScanPath` | `…/eraseCompleteScan` | `…\eraseCompleteScan` |

Full paths rather than a shared root plus separator constants. The two are genuinely different strings, and a root-plus-join scheme invites the manager — which is compiled for Linux but emits Windows pod specs — to build a Windows path with `filepath.Join` and get forward slashes that only fail at runtime, on the node.

**Linux values are byte-for-byte what they were.** No caller changes: every consumer already refers to the constants by name.

## Testing

`TestHandoffEndpointsFitInSunPath` (Windows-only) asserts each constant is an absolute Windows path and fits in `sun_path`. Those endpoints are Unix domain sockets on Windows, so the constants themselves are subject to the limit — not just the guard inside `listen`, and the constants are what a real deployment uses. Longest is 49 bytes against the 107 ceiling, so there is room, but the check means a future path change can't silently overrun it.

Verified: `GOOS=linux` and `GOOS=windows` build + vet clean, `golangci-lint` clean on both, `go test ./pkg/...` green.

## Coordination with #1236

#1236 adds `LinuxSharedDataPath` / `WindowsSharedDataPath` for the **manager**, which needs both values at once because it runs on Linux and emits specs for Windows. This PR is the **worker** half, which needs only its own and gets it at compile time.

The two are complementary but touch the same `const` block in `pkg/utils/utils.go`, so whichever merges second needs a small rebase. Happy to go in either order — if #1236 lands first I'll rebase this onto it.

## Follow-up, not in this PR

The remover's single five-minute context still spans `ListImages` + `ListContainers` + *every* `DeleteImage`. Measured on an AKS Windows node, one image takes 15–74s to delete and a freshly created node carried 30 unused images, so that budget is exhausted long before the work is. Sending that separately.


ashnamehrotra (approved):

pkg/utils/platform_windows.go
```diff
@@ -14,6 +14,16 @@ import (
 // a HostProcess pod reaches the pipe directly.
 const CRIPath = `npipe://./pipe/containerd-containerd`
 
+// The handoff endpoints, rooted where the manager mounts the shared-data volume
+// on this platform. These are Unix domain sockets, so the whole path has to fit
+// in sun_path -- see maxSocketPath in handoff_windows.go.
+const (
+	ScanErasePath            = `C:\run\eraser.sh\shared-data\scanErase`
```
sozercan, line 27: Could we use sandbox-relative paths here instead of hard-coding `C:\run`? [Kubernetes documents](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/#volume-mounts) that containerd 1.6 and fallback hosts expose these mounts beneath `%CONTAINER_SANDBOX_MOUNT_POINT%`. Would the current paths miss the shared volume and break every worker handoff? If those hosts are intentionally unsupported, could we enforce and document that requirement?

charleswool: Good point and looks like the second option you suggested is the safer way. Taken your alternative and made the requirement explicit rather than implicit:

```go
// The literal path is what a HostProcess container sees on containerd 1.7,
// which is what AKS runs and what these were measured against. Kubernetes
// documents that containerd 1.6 and the fallback hosts instead expose volume
// mounts beneath CONTAINER_SANDBOX_MOUNT_POINT, where none of this exists --
// so the worker handoff requires containerd 1.7 or newer.
```
One more things I can add if you feel needed. A startup check: the worker could confirm the shared-data directory exists before binding and fail with that sentence instead of a bind error. Cheap, and turns an unsupported host into a clear message rather than a confusing one.

Let me know if this is necessary

ashnamehrotra: thanks, I do not think the startup check is necessary

ashnamehrotra:
@charleswool  looks good to me, can you rebase PR and resolve merge conflicts?

PR #1272, opened 2026-09-02
fix: bump grpc to v1.83.1 to clear CVE-2026-84304 (repo-wide Trivy failure)

charleswool:
Trivy has started failing on every PR in the repo, including ones that change nothing relevant. This clears it.

## What is failing

```text
Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 1, CRITICAL: 0)

│        Library         │ Vulnerability  │ Severity │ Status │ Installed Version │ Fixed Version │
│ google.golang.org/grpc │ CVE-2026-84304 │ HIGH     │ fixed  │ v1.82.2           │ 1.83.1        │
```

## Why it appeared without anything changing

`v1.82.2` arrived on `main` via #1243 and was fine at the time. Trivy downloads a fresh vulnerability DB on every run, so the failure started the moment this CVE landed in that DB, not when the dependency changed.

The timing is visible on #1237, where two runs of the same branch straddle it:

| Run | Commit | Time (UTC) | Trivy |
|---|---|---|---|
| `33578224232` | `96e06a4a` — merge of `main` | 01:09 | pass |
| `33586510250` | `61a59ab4` | 03:19 | **fail** |

`main` itself has not run `test` since 2026-09-01 17:44, so it is green only because nothing has re-run it. It would fail identically today.

## The change

`google.golang.org/grpc` `v1.82.2` → `v1.83.1`, plus what `go mod tidy` pulls in as a consequence.

I took **v1.83.1**, exactly the version Trivy names as fixed, rather than the newest patch. `v1.83.2` also clears the CVE but drags `golang.org/x/net`, `x/sys`, `x/term`, `x/text` and `x/tools` along with it — 26 changed lines in `go.mod` against 16 for `v1.83.1`. For a security bump that has to land quickly, the smaller blast radius seemed the better trade; dependabot will pick up `v1.83.2` on its own schedule.

The `otel` `v1.43.0` → `v1.44.0` and `genproto` bumps are not optional — they are required by `grpc v1.83.x` itself.

## Verification

- `go build ./...` clean for `GOOS=linux` and `GOOS=windows`
- `go test ./pkg/... ./controllers/... ./api/...` all green
- `golangci-lint run ./...` clean under `GOOS=linux`
- `go mod verify` — all modules verified

No source changes, only `go.mod` / `go.sum`.

Same shape as #1232, which cleared the previous repo-wide Trivy failure.


ashnamehrotra (approved):
LGTM

